PR #52, opened 2025-09-12
Feature est 1004 ability of uploading files to api requests

VitaliiPovoroznykHYS:


src/crew/services/graph/nodes/file_extractor_node.py
```diff
@@ -0,0 +1,136 @@
+from services.graph.nodes.python_node import PythonNode
+from models.request_models import PythonCodeData
+
+
+class FileContentExtractorNode(PythonNode):
```
ArturDrohomeretskyiHYS, line 5: Add type of Node:
`TYPE = "FILE_EXTRACTOR"`

than extend available node types in src\crew\models\state.py for StateHistoryItem.type

src/django_app/tables/models/graph_models.py
```diff
@@ -67,6 +67,20 @@ class Meta:
         ]
 
 
+class FileExtractorNode(BaseNode):
```
ArturDrohomeretskyiHYS, line 70: where is migrations for new model?

PR #86, opened 2025-09-23
fix (BE | django_app): add flow validation for file extractor node

VitaliiPovoroznykHYS:


src/django_app/tables/services/session_manager_service.py
```diff
@@ -122,6 +126,11 @@ def create_session_data(
 
         file_extractor_node_data_list: list[FileExtractorNodeData] = []
         for item in file_extractor_node_list:
+            if not item.input_map:
```
ArturDrohomeretskyiHYS: move validation logic to separate class e.g. FileExtractorValidator in src\django_app\tables\validators\

VitaliiPovoroznykHYS: done

PR #89, opened 2025-09-24
Feature est 609 support mcp servers as tools

MaksymYevstratievHYS:


src/crew/services/crew/proxy_tool_factory.py
```diff
@@ -1,31 +1,29 @@
 import time
 from typing import Any, Type
 from crewai.tools.base_tool import Tool
-from models.request_models import CodeResultData, CodeTaskData
-from services.redis_service import RedisService
 from models.response_models import ToolResponse
```
SergeyTokarevHYS, line 4: pls use pep8 guide for imports 
https://peps.python.org/pep-0008/#imports
Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.

PR #92, opened 2025-09-24
Feature EST-1070 default context field

ArturDrohomeretskyiHYS:


src/django_app/tables/validators/end_node_validator.py
```diff
@@ -0,0 +1,73 @@
...
+
+class EndNodeValidator:
+    def validate(self, graph_id: int) -> EndNode | None:
+        end_node = self._check_exists(graph_id)
+
+        if end_node is not None: 
+            self._validate_output_map(end_node, graph_id)
+
+            if not end_node.output_map:
+                end_node = self._set_default_output_map(end_node)
```
VitaliiPovoroznykHYS: Would be better to not make changes for the node in validation, and move `set_default_output_map` to EndNode model

PR #93, opened 2025-09-24
feat(BE | django_app): add filtering by query param 'collection_id' f…

ArturDrohomeretskyiHYS:
…or EP: collection_statuses/

### DIFF
--- a/src/django_app/tables/views/views.py
+++ b/src/django_app/tables/views/views.py
@@ -681,7 +681,20 @@ class InitRealtimeAPIView(APIView):
 
 
 class CollectionStatusAPIView(APIView):
+    @swagger_auto_schema(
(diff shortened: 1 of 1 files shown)


src/django_app/tables/views/views.py
```diff
@@ -681,7 +681,20 @@ def post(self, request):
...
+                "collection_id",
+                openapi.IN_QUERY,
+                description="Filter by collection_id",
+                type=openapi.TYPE_STRING,
+                required=False,
+            )
+        ]
+    )
     def get(self, request):
+        collection_id = request.query_params.get("collection_id")
```
VitaliiPovoroznykHYS: Better move filtering logic to django-filters

PR #123, opened 2025-10-03
Feat EST-926 knowledge quering

ArturDrohomeretskyiHYS:


src/crew/libraries/crewAI/src/crewai/task.py
```diff
@@ -304,8 +307,9 @@ def execute_sync(
     @property
     def key(self) -> str:
         description = self._original_description or self.description
+        knowledge_query = self._original_knowledge_query or self.knowledge_query
```
MaksymYevstratievHYS, line 310: Which has higher priority?

"a" or "b" == "a"
Maybe swap

ArturDrohomeretskyiHYS: The current order is fine as is, no change needed

MaksymYevstratievHYS (approved):

PR #126, opened 2025-10-06
fix (BE | django_app): fix tasks context import order

VitaliiPovoroznykHYS:


src/django_app/tables/services/import_services.py
```diff
@@ -273,8 +274,10 @@ def __init__(self):
 
     def create_task(self, task_data, crew):
         current_id = task_data.pop("id", None)
+        data = deepcopy(task_data)
```
MaksymYevstratievHYS, line 277: I don't like this, but OK.

MaksymYevstratievHYS (approved):

PR #140, opened 2025-10-27
Fix est 1267 error after merge

MaksymHorelchykHYS:
Fix 1267 by auto sync primary key ids in db to prevent duplicate key errors
AI models update: added is_custom flag to keep all custom models in db


src/django_app/tables/management/commands/fix_sequences.py
```diff
@@ -0,0 +1,36 @@
...
+                    continue
+
+                if meta.pk.get_internal_type() not in ("AutoField", "BigAutoField", "SmallAutoField"):
+                    continue
+
+                try:
+                    cursor.execute(f"""
+                        SELECT setval(
+                            pg_get_serial_sequence('"{meta.db_table}"', 'id'),
+                            (SELECT COALESCE(MAX(id), 1) FROM "{meta.db_table}")
```
ArturDrohomeretskyiHYS: When a table is empty, COALESCE(MAX(id), 1) sets the sequence value to 1, meaning the next insert will use id = 2. So the first instance in table will be with id 2
Consider using 0 instead of 1

src/django_app/tables/management/commands/upload_models.py
```diff
@@ -78,6 +78,8 @@ def upload_llm_models():
     models_by_provider = load_json_from_file(path)
     current_model_tuples = set()
 
+    LLMModel.objects.all().delete()
```
ArturDrohomeretskyiHYS: remove this line
firstly you delete all table instances (with customs)
and than delete with excluding custom instances (None)

src/django_app/tables/management/commands/upload_models.py
```diff
@@ -99,6 +101,8 @@ def upload_realtime_agent_models():
     models_by_provider = load_json_from_file(path)
     current_model_tuples = set()    
 
+    RealtimeModel.objects.all().delete()
```
ArturDrohomeretskyiHYS: remove this line
same error

src/django_app/tables/management/commands/upload_models.py
```diff
@@ -118,6 +122,8 @@ def upload_realtime_transcription_models():
     models_by_provider = load_json_from_file(path)
     current_model_tuples = set()    
 
+    RealtimeTranscriptionModel.objects.all().delete()
```
ArturDrohomeretskyiHYS: remove this line
same error

src/django_app/tables/management/commands/upload_models.py
```diff
@@ -137,6 +143,8 @@ def upload_embedding_models():
     models_by_provider = load_json_from_file(path)
     current_model_tuples = set()    
     
+    EmbeddingModel.objects.all().delete()
```
ArturDrohomeretskyiHYS: remove this line
same error

PR #177, opened 2025-11-19
Feature est 1269 webhook node

MaksymYevstratievHYS:
Added tests and added migration

### DIFF
--- a/src/.dockerignore
+++ b/src/.dockerignore
@@ -1,5 +1,5 @@
 ENV
 */.venv
 **/venv
-**/*venv*
+**/*venv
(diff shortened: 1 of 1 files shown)


src/.dockerignore
```diff
@@ -1,4 +1,5 @@
 ENV
 */.venv
 **/venv
+**/*venv*
```
ArturDrohomeretskyiHYS: Okay, but consider: using this pattern can unintentionally ignore any folder that happens to contain the word “venv” in its name. Even if it is important dir

PR #186, opened 2025-11-28
Fix est 1512 stop pending sessions

MaksymYevstratievHYS:


src/crew/services/graph/graph_session_manager_service.py
ArturDrohomeretskyiHYS, line 262: Sessions executed?

ArturDrohomeretskyiHYS (approved):

PR #187, opened 2025-12-01
fix(BE | sandbox): fix task creation

MaksymYevstratievHYS:
Only 1 coroutine was running


src/sandbox/main.py
```diff
@@ -13,20 +13,19 @@
...
+    base_venv_path=base_venv_path,
+)
 os.chdir("savefiles")
 
+redis_service = RedisService(host=redis_host, port=redis_port)
 
-async def main():
-    redis_service = RedisService(host=redis_host, port=redis_port)
-
+async def init():
```
ArturDrohomeretskyiHYS, line 24: Suggest to rename init() to main(). This will be consistent, like in knowledge service

PR #193, opened 2025-12-05
feat(BE): restore python code tool feature

MaksymYevstratievHYS:


src/django_app/tables/migrations/0128_crew_is_template_pythoncodetoolconfig_and_more.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("tables", "0127_fix_graphsessionmessage_cascade"),
+    ]
+
+    operations = [
+        migrations.AddField(
```
VitaliiPovoroznykHYS: This migration should not be here. It's from feature EST-1411 and should not be on any other branch

src/django_app/tables/models/crew_models.py
```diff
@@ -247,6 +262,7 @@ class Crew(AbstractDefaultFillableModel):
         blank=True,
         help_text="Float between 0.00 and 1.00 for knowledge",
     )
+    is_template = models.BooleanField(default=False)
```
VitaliiPovoroznykHYS: Same here. This is from 1411

PR #196, opened 2025-12-08
Feat est 1148 web node

MaksymHorelchykHYS:


src/crew/services/graph/nodes/web_scraper_knowledge_node.py
```diff
@@ -0,0 +1,218 @@
...
+            return result
+        time.sleep(wait_interval)
+        waited += wait_interval
+        collection = get_collection_by_name(collection_name)
+    return collection
+
+def wait_completed(collection_name: str):
+    collection = get_collection_by_name(collection_name)
+    url = f"{{API_BASE}}/collection_statuses/?collection_id={{collection['collection_id']}}"
+    while True:
```
VitaliiPovoroznykHYS, line 115: Add some timeout for this loop. Otherwise we might never get status "completed" and there will be just infinite loop

src/crew/services/graph/nodes/web_scraper_knowledge_node.py
```diff
@@ -0,0 +1,218 @@
...
+
+    existing_collection = if_exists(collection_name, time_to_expired, urls)
+    if existing_collection:
+        return existing_collection
+
+    base_dir = prepare_save_folder(collection_name)
+    scraped_contents = asyncio.run(scrape_all_urls(urls))
+    file_names = [save_scraped_file(base_dir, url, content) for url, content in zip(urls, scraped_contents)]
+    response = create_collection(collection_name, file_names, embedder, urls)
+    wait_completed(collection_name)
```
VitaliiPovoroznykHYS, line 216: What if collection was not completed or got some error/warning status?

VitaliiPovoroznykHYS (changes requested):

src/crew/services/graph/nodes/web_scraper_knowledge_node.py
MaksymHorelchykHYS: resolve

src/crew/services/graph/nodes/web_scraper_knowledge_node.py
MaksymHorelchykHYS: resolve

PR #211, opened 2026-01-13
feat(EST-1749-BE): add token usage to session messages

VitaliiPovoroznykHYS:


src/django_app/tables/services/redis_pubsub.py
```diff
@@ -131,16 +132,46 @@ def _save_organization_variables(self, session: Session, data: dict):
...
-                )
+
+                created_objects = model.objects.bulk_create(data, ignore_conflicts=True)
                 logger.debug(
                     f"{model.__name__} updated with {len(created_objects)}/{len(data)} entities"
                 )
         except IntegrityError as e:
             logger.error(f"Failed to save {model.__name__}: {e}")
 
+    def _calculate_total_token_usage(self, session_id):
```
ArturDrohomeretskyiHYS, line 141: Suggestion for future refactoring:

Consider moving token calculation logic to the crew container.
1. This would improve separation of concerns
2. Сalculate tokens at their source (where token usage data is initialized)
3. Eliminate the dependency on Redis cache for per-node token aggregation

PR #216, opened 2026-01-15
fix!: fix ports

MaksymYevstratievHYS:
Hotfix for managing ports from .env file

### DIFF
--- a/src/docker-compose.yaml
+++ b/src/docker-compose.yaml
@@ -7,7 +7,7 @@ services:
     command: ./entrypoint.sh
     container_name: django_app
     ports:
-      - "${DJANGO_PORT}:${DJANGO_PORT}"
+      - "${DJANGO_PORT:-8000}:${DJANGO_PORT:-8000}"
     environment:
       - CODE_RESULT_CHANNEL=code_results
       - CODE_EXEC_TASK_CHANNEL=code_exec_tasks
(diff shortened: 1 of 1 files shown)


src/docker-compose.yaml
```diff
@@ -50,7 +49,8 @@ services:
         condition: service_healthy
   
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8000/ht/"]
+      test: >
+        sh -c 'curl -f http://localhost:${DJANGO_PORT}/ht/ || exit 1'
```
SergeyTokarevHYS: what if django will be hosted in another mashine?

src/docker-compose.yaml
```diff
@@ -62,12 +62,12 @@ services:
       context: ./realtime
       dockerfile: Dockerfile.realtime
     ports:
-    - "8050:8050"
+      - "${REALTIME_PORT}:${REALTIME_PORT}"
     environment:
       - KNOWLEDGE_SEARCH_GET_CHANNEL=knowledge:search:get
```
SergeyTokarevHYS, line 70: you can use merge keys for thos
x-shared-env: & shared-env
 environment:
      # Use the shared environment variables.
      <<: *shared-env

src/realtime/run_server.py
```diff
@@ -10,14 +11,14 @@
 else:
     load_dotenv(find_dotenv(".env"))
 
-
+PORT = os.environ.get("REALTIME_PORT", 8050)
```
SergeyTokarevHYS: pls create file settings py 
and  use pydantic settings
https://docs.pydantic.dev/latest/concepts/pydantic_settings/#usage

src/realtime/run_server.py
```diff
@@ -10,14 +11,14 @@
...
 def main():
     """Run the FastAPI server with uvicorn."""
     uvicorn.run(
         "api.main:app",
         host="0.0.0.0",
-        port=8050,
+        port=PORT,
         reload=False,
         reload_dirs=["src"],
         workers=1,
```
SergeyTokarevHYS: workers should be more than one and be also configurable

src/.env
```diff
@@ -33,18 +29,21 @@ ALLOWED_HOSTS=0.0.0.0,127.0.0.1,localhost,django_app
 GUNICORN_WORKERS=1
 GUNICORN_THREADS=4
 MEDIA_ROOT=/app/media
+DB_NAME=crew
 DJANGO_PORT=8000
```
SergeyTokarevHYS, line 35: we not should use standard ports for this

src/realtime/run_server.py
```diff
@@ -10,14 +11,14 @@
...
 
 def main():
     """Run the FastAPI server with uvicorn."""
     uvicorn.run(
         "api.main:app",
         host="0.0.0.0",
-        port=8050,
+        port=PORT,
         reload=False,
         reload_dirs=["src"],
```
SergeyTokarevHYS: reload_dirs=["src"], should be added only if DEBUG=TRUE

src/docker-compose.yaml
```diff
@@ -306,7 +308,7 @@ services:
       dockerfile: Dockerfile.fe
     container_name: frontend
```
SergeyTokarevHYS: we probably should rename it on smt like epicstaff-frontend

src/debug.env
```diff
@@ -1,4 +1,3 @@
-DB_NAME=crew
 # ===== SHARED =====
 # Redis Configuration
 REDIS_HOST=127.0.0.1
```
SergeyTokarevHYS, line 3: we also need secure our redis and fill the password for it

src/.env
MaksymYevstratievHYS: it is configurable

PR #228, opened 2026-01-26
Feature/est 1798 warnings msg for flow/be

YaroslavArtiushenkoHYS:


src/django_app/tables/models/session_models.py
```diff
@@ -114,3 +115,11 @@ class TaskSessionMessage(CrewSessionMessage):
     expected_output = models.TextField(blank=True, default="")
     raw = models.TextField(blank=True, default="")
     agent = models.TextField(blank=True, default="")
+
+
+class SessionWarningMessage(models.Model):
+    session = models.ForeignKey(
```
VitaliiPovoroznykHYS: Using OneToOne field is better, cause we will never have more than one record about warnings for single session, cause we store all warnings inside JSON field here

src/django_app/tables/views/views.py
```diff
@@ -328,8 +361,17 @@ def post(self, request):
...
+                warnings_exist = True
+                SessionWarningMessage.objects.create(
+                    session_id=session_id, messages=warning_messages
+                )
+            else:
+                warnings_exist = False
+
             return Response(
-                data={"session_id": session_id}, status=status.HTTP_201_CREATED
+                data={"session_id": session_id, "warnings_exist": warnings_exist},
```
VitaliiPovoroznykHYS: In this solution we don't need to pass "warnings_exist" flag to frontend on run_session, cause frontend will just use separate endpoint to see if there are any warnings anyway

src/django_app/tables/views/views.py
```diff
@@ -267,6 +293,12 @@ def post(self, request):
             graph__id=graph_id
         ).first()
 
+        if graph_organization:
+            if not username and graph_organization.user_variables:
+                warning_messages.append(
+                    "Current flow have user-persistent variables, but no user provided"
```
VitaliiPovoroznykHYS: Create enums.py on top of "tables" app and create SessionWarningEnum or something like that. Let's have all warnings there in one place

VitaliiPovoroznykHYS: And also let's change the message to something like: "This flow has user persistent variables, but since user was not provided, this run will use your default flow variables"

src/django_app/tables/migrations/0136_alter_sessionwarningmessage_session.py
VitaliiPovoroznykHYS, line 1: Okay, let's not create unnecessary migrations here. Remove your 0135 and 0136 them first locally, merge current main into this branch, and create new single migration after you got all migrations from main branch

src/django_app/tables/views/views.py
```diff
@@ -76,6 +77,8 @@
     RegisterTelegramTriggerSerializer,
 )
 
+from tables.models.session_models import SessionWarningMessage
```
VitaliiPovoroznykHYS: Also fix this import to `from tables.models import SessionWarningMessage` and do that somewhere on line 54, where all other models are being imported. This works, since we have all models imported in __init__.py inside models directory, so we can use import like this one

PR #253, opened 2026-02-04
fix: fix NGINX ssl enable and frontend config

MaksymYevstratievHYS:


src/.env
```diff
@@ -82,6 +82,9 @@ WEBHOOK_PORT=8009
 
 # frontend
 FRONTEND_PORT=4200
+API_URL=http://127.0.0.1/api/
+REALTIME_API_URL=http://127.0.0.1/realtime/
 
 # nginx
 DOMAIN_NAME="localhost"
+SSL_ENABLE=#
```
ArturDrohomeretskyiHYS, line 93: not robust solution 

can use two different .conf files for nginx with ssl and without as substrings

nginx-ssl.conf

```
server {
    listen 443 ssl;
    server_name ${DOMAIN_NAME};

    ssl_certificate /etc/nginx/certs/fullchain.pem;
    ssl_certificate_key /etc/nginx/certs/privkey.pem;

    ssl_protocols TLSv1.2 TLSv1.3;

    location / {
        proxy_pass http://app;
    }
}
```

nginx-no-ssl.conf

```
server {
    listen 80;
    server_name ${DOMAIN_NAME};

    location / {
        proxy_pass http://app;
    }
}

```

and substitute in entrypoint based on bool param SSL_ENABLE:

```
if [ "$SSL_ENABLE" = "true" ]; then
  envsubst < nginx-ssl.conf > /etc/nginx/conf.d/default.conf.template
else
  envsubst < nginx-no-ssl.conf > /etc/nginx/conf.d/default.conf.template
fi
```

PR #259, opened 2026-02-06
Feature/est 1926 imp auth

SergeyTokarevHYS:


frontend/src/app/pages/setup-page/setup-page.component.html
```diff
@@ -0,0 +1,43 @@
+<div class="setup-page">
+  <div class="setup-card" *ngIf="step === 'form'">
+    <h1>Initial Setup</h1>
+    <p>Create the first admin user to get started.</p>
+
+    <form [formGroup]="form" (ngSubmit)="submit()">
+      <label>
+        Username
+        <input type="text" formControlName="username" autocomplete="username" />
```
BohdanSydirHYS: you can use shared components that have the necessary styling

Check: app-custom-input

frontend/src/app/pages/setup-page/setup-page.component.html
```diff
@@ -0,0 +1,43 @@
...
+        Password
+        <input type="password" formControlName="password" autocomplete="new-password" />
+      </label>
+
+      <label>
+        Email (optional)
+        <input type="email" formControlName="email" autocomplete="email" />
+      </label>
+
+      <button type="submit" [disabled]="form.invalid || loading">
```
BohdanSydirHYS: you can use shared components that have the necessary styling, and input params

Check: app-button

frontend/src/app/pages/setup-page/setup-page.component.scss
```diff
@@ -0,0 +1,81 @@
...
+  border: 1px dashed #94a3b8;
+  border-radius: 8px;
+  word-break: break-all;
+}
+
+.error {
+  margin-top: 12px;
+  color: #dc2626;
+  font-size: 13px;
+}
```
BohdanSydirHYS: Use BEM (block__element--modifier) methodology for styling.

frontend/src/app/pages/login-page/login-page.component.scss
```diff
@@ -0,0 +1,60 @@
...
+button:disabled {
+  background: #94a3b8;
+  cursor: not-allowed;
+}
+
+.error {
+  margin-top: 12px;
+  color: #dc2626;
+  font-size: 13px;
+}
```
BohdanSydirHYS: Use BEM (block__element--modifier) methodology for styling.

frontend/src/app/pages/login-page/login-page.component.html
```diff
@@ -0,0 +1,23 @@
+<div class="login-page">
+  <div class="login-card">
+    <h1>Sign In</h1>
+
+    <form [formGroup]="form" (ngSubmit)="submit()">
+      <label>
+        Username
+        <input type="text" formControlName="username" autocomplete="username" />
+      </label>
```
BohdanSydirHYS: You can use shared components instead of creating your own, example:

`<app-custom-input
                        label="Username"
                        tooltipText="Username"
                        formControlName="username"
                        placeholder="Enter username"
                        [errorMessage]="error"
                    ></app-custom-input>`

BohdanSydirHYS (commented):
Use shared components instead of creating new; Use BEM methodology for styling

frontend/src/app/pages/login-page/login-page.component.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  form = new FormGroup({
+    username: new FormControl('', { nonNullable: true, validators: [Validators.required] }),
+    password: new FormControl('', { nonNullable: true, validators: [Validators.required] }),
+  });
+
+  error: string | null = null;
+  loading = false;
+
+  async submit(): Promise<void> {
```
BohdanSydirHYS: async is not necessary

Promise<void> is never returning and not necessary

src/django_app/tables/views/auth_views.py
ArturDrohomeretskyiHYS, line 1: move all swagger schema to dedicated path:
src\django_app\tables\swagger_schemas

src/django_app/tables/views/auth_views.py
```diff
@@ -0,0 +1,311 @@
...
+            {
+                "access": str(refresh.access_token),
+                "refresh": str(refresh),
+                "api_key": raw_key,
+            },
+            status=status.HTTP_201_CREATED,
+        )
+
+
+class FirstSetupView(APIView):
```
ArturDrohomeretskyiHYS, line 250: Race condition / no atomicity in FirstSetupView.post and ResetUserView.post
Both views check state then mutate without transaction.atomic() or select_for_update(). Two concurrent POST /api/auth/first-setup/ requests could both pass the user_model.objects.exists() check and create duplicate superusers. Same for reset-user. Wrap these in transaction.atomic().

src/django_app/tables/views/auth_views.py
```diff
@@ -0,0 +1,311 @@
...
+            )
+        return Response(
+            {
+                "access_token": serializer.validated_data["access"],
+                "token_type": "bearer",
+            }
+        )
+
+
+class ResetUserView(APIView):
```
ArturDrohomeretskyiHYS, line 202: Race condition / no atomicity

src/django_app/django_app/settings.py
ArturDrohomeretskyiHYS, line 94: Future improwement:
CorsMiddleware is placed last in MIDDLEWARE
CorsMiddleware should be placed as high as possible (before CommonMiddleware) so CORS headers are set on error responses too. Currently it's last, which means CORS preflight failures may not get proper headers.

frontend/src/app/visual-programming/components/node-panels/decision-table-node-panel/decision-table-grid/cell-editors/expression-editor/autocomplete-overlay/.DS_Store
ArturDrohomeretskyiHYS, line 1: must be added to .gitignore

frontend/src/app/visual-programming/components/node-panels/decision-table-node-panel/decision-table-grid/cell-editors/expression-editor/.DS_Store
ArturDrohomeretskyiHYS, line 1: must be added to .gitignore

src/django_app/entrypoint.sh
```diff
@@ -19,6 +19,59 @@ python manage.py fix_sequences
...
+      GENERATED_ADMIN_PASSWORD=$(python - <<'PY'
+import secrets
+print(secrets.token_urlsafe(18))
+PY
+)
+      export DJANGO_ADMIN_PASSWORD="${GENERATED_ADMIN_PASSWORD}"
+    fi
+    echo "Ensuring default Django admin user exists..."
+    python manage.py shell -c "
+from django.contrib.auth import get_user_model
```
ArturDrohomeretskyiHYS, line 36: refactor later for RBAC

src/django_app/tables/management/commands/reset_user.py
ArturDrohomeretskyiHYS, line 1: refactor for RBAC
get_user_model

src/django_app/tables/views/auth_views.py
```diff
@@ -0,0 +1,311 @@
+from rest_framework.permissions import IsAuthenticated, AllowAny
+from rest_framework.response import Response
+from rest_framework.views import APIView
+from rest_framework import status, serializers as drf_serializers
+from rest_framework_simplejwt.tokens import AccessToken
+from rest_framework_simplejwt.exceptions import TokenError
+from rest_framework_simplejwt.tokens import RefreshToken
+from rest_framework_simplejwt.serializers import TokenObtainPairSerializer
+from django.contrib.auth import get_user_model
```
ArturDrohomeretskyiHYS, line 9: refactor for RBAC

ArturDrohomeretskyiHYS:
Technical Debt: Auth layer requires structural refactoring for RBAC Stories
The auth implementation in this PR works for the current single-user, flat-permission model, but it has several structural assumptions that are incompatible with the upcoming RBAC schema. Flagging these now so they're tracked as known debt and addressed during RBAC Stories rather than discovered mid-sprint.
1. username is hardcoded across the entire auth surface — the new User model won't have it.
The custom User model (RBAC Story) uses email as USERNAME_FIELD and drops username entirely. FirstSetupView, ResetUserView, SwaggerTokenView, AuthMeView, TokenIntrospectView, the reset_user management command, all tests, and the .env file all reference username. These will raise TypeError or return None after the model swap. This affects every auth endpoint without exception.
2. request.user = ApiKey makes RBAC permission checks impossible.
JwtOrApiKeyAuthentication._authenticate_api_key returns (ApiKey_instance, {"scopes": ...}), so request.user is an ApiKey object, not a User. The RBAC model requires resolving every request to User → OrganizationUser → Role → RolePermission. None of that chain works when the user object is an API key. ApiKey needs a created_by FK to User (and ideally an org FK), and the auth backend needs to return the actual owning user so downstream permission checks have something to work with. ApiKey.is_authenticated = True (unconditional) compounds this by making IsAuthenticated pass for all keys regardless of context.
3. HasScope (string-based) is a different permission system than RBAC (bitmask-based).
permissions.py checks string scopes like ["introspect"]. The new model uses integer bitmasks on RolePermission with resource_type enum. These have no shared interface. HasScope will become dead code after RBAC Story and needs to be replaced with a permission class that evaluates OrganizationUser.role → RolePermission → bitmask.
4. Setup/reset views have no transaction safety — RBAC makes this worse.
FirstSetupView.post and ResetUserView.post perform multiple DB mutations (delete users, delete keys, create superuser, create key) without transaction.atomic(). After RBAC Story setup will also need to create an Organization, OrganizationUser, and assign a role. A partial failure midway leaves the system unrecoverable — no users, no keys, no way to authenticate. Same applies to the reset_user management command.
5. No org context exists anywhere in the auth layer.
The RBAC schema is multi-org (OrganizationUser with one role per user per org). Currently there's no org ID in JWT claims, no org resolution on requests, and no org scoping on any endpoint. JWT generation (TokenObtainPairView) will need to be customized to embed org claims, and the auth backend will need to resolve which org a request is scoped to. This is a prerequisite for Stories 4 and 5.
6. /auth/me/ returns wrong data for API key callers and lacks RBAC fields.
When request.user is an ApiKey, the endpoint returns the key's PK as id and None for everything else. After RBAC, this endpoint will be expected to return org membership, role, and effective permissions — the frontend uses it to determine what the user can see and do.

frontend/src/app/visual-programming/components/node-panels/decision-table-node-panel/decision-table-grid/cell-editors/expression-editor/autocomplete-overlay/.DS_Store
ArturDrohomeretskyiHYS: @BohdanSydirHYS

frontend/src/app/visual-programming/components/node-panels/decision-table-node-panel/decision-table-grid/cell-editors/expression-editor/.DS_Store
ArturDrohomeretskyiHYS: @BohdanSydirHYS

src/django_app/tables/views/auth_views.py
ArturDrohomeretskyiHYS: @MaksymYevstratievHYS

PR #288, opened 2026-02-17
Feature/est 1984 add meta fields

MaksymYevstratievHYS:


src/django_app/tables/models/graph_models.py
```diff
@@ -16,8 +18,32 @@ class Graph(models.Model):
         default=False, help_text="If 'True' -> use variables from last session."
     )
 
+class MetadataMixin(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+    content_hash = models.CharField(max_length=64, editable=False, null=True)
+    metadata = models.JSONField(default=dict, blank=True)
```
VitaliiPovoroznykHYS: I think it's better to move metadata field to BaseNode

src/django_app/tables/models/graph_models.py
```diff
@@ -16,8 +18,32 @@ class Graph(models.Model):
         default=False, help_text="If 'True' -> use variables from last session."
     )
 
+class MetadataMixin(models.Model):
```
VitaliiPovoroznykHYS: It would be better to separate this class into 2:
- TimestampModelMixin: contains `created_at`, `updated_at` fields
- HashableModelMixin: contains `content_hash`, `generate_hash()` and `get_hash_exluded_fields()`

In this way we can reuse timestamps everywhere in the project and add ability to have hash for any model as well. Also "Model" in mixin name sounds better so we clearly understand that we should use this with models only.

`get_hash_exluded_fields()` - should return a list of fields that will not be used in hash and should be overridden by child classes

src/django_app/tables/serializers/base_serializer.py
```diff
@@ -0,0 +1,16 @@
+from rest_framework import serializers
+
+class BaseMetadataSerializer(serializers.ModelSerializer):
```
VitaliiPovoroznykHYS: Same idea with separation for Timestamp and Hashable

src/django_app/tables/models/graph_models.py
VitaliiPovoroznykHYS: For Timestamps, maybe it's even better to not use Mixin, but create some BaseTimestampModel and keep only Hashas as mixin. It's up to you to decide here

src/django_app/tables/models/graph_models.py
```diff
@@ -18,32 +21,8 @@ class Graph(models.Model):
...
-        }
-        data_string = json.dumps(data, sort_keys=True).encode('utf-8')
-        return hashlib.sha256(data_string).hexdigest()
-
-    def save(self, *args, **kwargs):
-        self.content_hash = self.generate_hash()
-        super().save(*args, **kwargs)
 
-class BaseNode(MetadataMixin, models.Model):
+class BaseNode(BaseGraphEntity, models.Model):
```
VitaliiPovoroznykHYS, line 25: We don't really need models.Model here, cause it's already in parent class

PR #306, opened 2026-02-20
Feature/est 893 show chunks/release candidate

ArturDrohomeretskyiHYS:


src/django_app/tables/views/knowledge_views/naive_rag_views.py
```diff
@@ -592,18 +601,216 @@ class NaiveRagChunkViewSet(ReadOnlyModelViewSet):
 
 
 class ProcessNaiveRagDocumentChunkingView(APIView):
```
MaksymYevstratievHYS, line 603: Move all business logic into service, keep views as thin as possible

src/knowledge/storage/naive_rag_storage.py
```diff
@@ -196,10 +198,13 @@ def save_document_chunks(
...
                     naive_rag_document_config_id=naive_rag_document_config_id,
-                    text=chunk_text,
+                    text=chunk_data.text,
                     chunk_index=idx,
+                    token_count=chunk_data.token_count,
+                    overlap_start_index=chunk_data.overlap_start_index,
+                    overlap_end_index=chunk_data.overlap_end_index,
                 )
-                for idx, chunk_text in enumerate(chunk_list)
+                for idx, chunk_data in enumerate(chunk_list, start=1)
```
MaksymYevstratievHYS, line 207: why start=1?

ArturDrohomeretskyiHYS: User friendly representation in UI. So the first chunk will be with index 1 instead of 0

PR #307, opened 2026-02-22
Code Agent Node

yuriwa:
## Code Agent Node

A new node type for EpicStaff that connects flows to an AI coding agent powered by OpenCode. It enables conversational, tool-using LLM interactions with real-time streaming — ideal for coding tasks, system administration, and complex multi-step reasoning within a flow.

### Key Features

- **Conversational sessions** — maintains persistent OpenCode sessions per `chat_id`, so follow-up messages continue the same conversation with full context
- **Real-time streaming** — polls OpenCode at 50ms intervals, streaming reasoning and tool call progress to the flow and to user-defined callbacks
- **Stream handler callbacks** — user-defined Python code (`on_stream_start`, `on_chunk`, `on_complete`) runs in the sandbox, enabling integrations like live Google Chat bubble updates
- **Handler state persistence** — return values from callbacks persist across calls (e.g., a GChat message ID created in `on_chunk` is available in `on_complete` to update the same message)
- **Fire-and-forget execution** — handler callbacks run as chained async tasks so they never block the main polling loop
- **Graceful stop** — on session stop, cancels in-flight work, notifies the user via `on_complete`, and releases the OpenCode instance if no other sessions need it
- **Instance lifecycle management** — OpenCode instances are provisioned per LLM config via the Instance Manager, with idle reaping and stop-on-last-user cleanup

### Configuration

| Field | Description |
|---|---|
| **LLM Config** | Which model to use (e.g., GPT-5.1) |
| **Agent Mode** | `code` or `plan` |
| **System Prompt** | Injected on first message in a session |
| **Stream Handler Code** | Python callbacks for real-time integrations |
| **Libraries** | pip packages needed by the handler (e.g., `google-auth`, `httpx[http2]`) |
| **Polling Interval** | Base poll rate in ms (actual rate is 1/10th for responsiveness) |

### Architecture

- **Backend**: Django model, serializer, viewset, migration, session manager integration
- **Frontend**: Node panel with Monaco editors for system prompt and stream handler, port definitions, save-graph integration
- **Runtime**: `CodeAgentNode` in crew, `instance_manager.py` in the code container, `Dockerfile.code` + `start.sh`
- **Skills**: EpicStaff CLI tools baked into OpenCode instances so the agent can query and manage flows, projects, tools, and sessions


docs/code_agent.md
```diff
@@ -0,0 +1,336 @@
...
+Use it when you need a flow to:
+- Run an autonomous coding agent that can execute commands, read/write files
+- Stream real-time progress (reasoning + tool calls) to external platforms (Google Chat, Telegram, etc.)
+- Maintain persistent conversations across multiple flow triggers (same `chat_id` = same session)
+
+---
+
+## Quick Start
+
+1. **Add a Code Agent node** to your flow in the visual editor
```
SergeyTokarevHYS, line 16: it cannot be added from UI

PR #309, opened 2026-02-23
fix(BE): removed request_models field from llm config

MaksymYevstratievHYS:


AndreyKrasnoyarskiy:
Merged EST-3149 instead

PR #348, opened 2026-03-11
feat(EST-2004-BE): implemented hash validation

MaksymYevstratievHYS:


src/django_app/tables/models/base_models.py
ArturDrohomeretskyiHYS, line 188: It is hard to debug nested inheritance

src/django_app/tables/models/graph_models.py
ArturDrohomeretskyiHYS, line 32: It is hard to debug nested inheritance

ArturDrohomeretskyiHYS, line 59: It is hard to understand inheritance logic across all Node models

PR #359, opened 2026-03-14
Feature/epic chat yuri streaming rc

ArturDrohomeretskyiHYS:
DRAFT PR, don't merge


src/.env
ArturDrohomeretskyiHYS, line 1: revert

src/docker-compose.yaml
```diff
@@ -346,9 +375,12 @@ volumes:
     external: true
   crew_pgdata:
     external: true
+  crew_config:
+    external: true
   media_data:
     external: true
   redis_data: {}
+  code_instances:
```
ArturDrohomeretskyiHYS, line 383: docker compose down will delete all data in this volume
Check if we need such a behavior. Volume code_instances will be empty for every new launch

PR #412, opened 2026-04-01
Feature/est 2029 schedule trigger node

ViraKalenykHYS:
**Summary**
Add a new Schedule Trigger node to the visual programming editor so flows can run automatically based on time-based configuration.

**Frontend scope**
- add the Schedule Trigger node to the canvas
- create the configuration panel in the right sidebar
- implement UI for start/end date configuration
- implement repeat frequency and custom CRON inputs
- add iterations and active window controls
- add frontend validation for schedule-related inputs

**Current status**
Work in progress.
At the moment, this PR contains the frontend visual and configuration part only.
Backend support, persistence, and integration with the API are not included yet and depend on the corresponding backend implementation.

**Definition of Done**
**Backend**
- `ScheduleTriggerNode` model is created and migrated
- scheduler service is implemented and processes time-based triggers correctly
- executions start a new `GraphSession` via the background worker
- start/end dates and iteration limits are respected
- graph import/export supports the new node

**Frontend**
- "Schedule Trigger" node is available in the tools menu and can be dropped onto the canvas
- the configuration panel includes inputs for Start/End dates, Iterations, Custom Duration, and Repeat/CRON logic
- saving the graph persists the schedule configuration to the API
- UI validation is implemented


ElenaPrygunHYS:
At the moment, there are no comments.
By the way, when creating new components, you don’t need to add standalone: true, as it’s the default in the new syntax.

PR #434, opened 2026-04-03
Feature/est 2334 file manager/release candidate

VitaliiPovoroznykHYS:
This PR provides new feature "File Manager" into our system. 

- There will be ability to store files both locally and any other S3 compatible storage. 
- MinIO is part of the requirements
- Users will be able to do most of the operations with files that can be done in common file systems
- There will be some limitations, like archives cannot be uploaded directly, when someone uploads an archive - it will be extracted and only its content will be in our storage. We also will not allow any kind of executable files to be stored
- After uploading files to file manager, users will be able to attach files to any flow
- To use files in any flow there will be separate SDK for python nodes to manage storage operations
- Each session that has files should have ability to download/preview them after a run


ElenaPrygunHYS:
Reviewed the existing FE commits - no comments so far. However, as discussed in our private conversation, it would be better to use a consistent new syntax (input/output instead of decorators) across the component.

ElenaPrygunHYS:
Frontend changes have been reviewed

ElenaPrygunHYS (approved):

ArturDrohomeretskyiHYS:
squash migrations pls

ArturDrohomeretskyiHYS (approved):

PR #438, opened 2026-04-06
Feature est 1165 openinterpreter2 as built in tool

PavloNesterukHYS:


VitaliiPovoroznykHYS:
Seems like we are not going to use open-interpreter anymore dues to AGPL license, so we will move this to separate repo

PR #439, opened 2026-04-07
Feature/EST-1965 Flow session window design

romanmasliak-byte:
This PR changes flow messages view directly in flow session window and some action buttons logic

Existing Status Filter dropdown app-flow-session-status-filter-dropdown was moved to other component inside the table according to Svitlana design


ElenaPrygunHYS:
@romanmasliak-byte , No comments at the moment. The only note is that when creating new components, it would be better to use the new syntax (signals, e.g. input()/output() instead of @Input/ @Output). In this case, it’s fine to keep the existing approach, since the changes were made to components that already use the older syntax.

ElenaPrygunHYS (approved):

PR #440, opened 2026-04-08
Feature / EST-2488 Import / Export backward compatibility

PavloNesterukHYS:
Story [EST-2488](https://hys-enterprise.atlassian.net/browse/EST-2488) 

As a user, I want a mechanism that can handle compatibility between my old and new exports. 

That would require:
- Adding a constant that shows current version
- Marking each import with that version
- Adding mechanism that could convert old exports to the current one in the future (similar to db migrations)


src/django_app/tables/import_export/services/export_service.py
```diff
@@ -3,11 +3,10 @@
 
 from tables.import_export.registry import EntityRegistry
 from tables.import_export.enums import EntityType
-from tables.import_export.constants import MAIN_ENTITY_KEY
+from tables.import_export.constants import MAIN_ENTITY_KEY, CURRENT_VERSION
```
VitaliiPovoroznykHYS: Let's rename this to IMPORT_VERSION, this sounds more clear

VitaliiPovoroznykHYS (changes requested):
Also update docs in root for this functionality briefly

VitaliiPovoroznykHYS (approved):

PR #448, opened 2026-04-14
Feat/EST-1761 python node code test case/fe

KarolinaKudrytskaHYS:
Added a test mode to Python Node. Works exclusively in expanded panel view.
A toggle has been implemented to enable or disable test mode.
A separate input list for enabled test mode, which allows you to assign any values to variables.
Displays the result as a terminal within the code editor.


src/django_app/tables/views/views.py
```diff
@@ -1020,3 +1024,97 @@ def post(self, request):
         webhook_trigger_service = WebhookTriggerService()
         webhook_trigger_service.register_webhooks()
         return Response(status=status.HTTP_200_OK)
+
+
+class PythonNodeLastTestInputView(APIView):
+    @swagger_auto_schema(
```
VitaliiPovoroznykHYS: Move this schema to src\django_app\tables\swagger_schemas

src/django_app/tables/views/views.py
```diff
@@ -1020,3 +1024,97 @@ def post(self, request):
...
+            404: openapi.Response(description="PythonNode not found"),
+        },
+    )
+    def get(self, request, pk):
+        try:
+            python_node = PythonNode.objects.get(pk=pk)
+        except PythonNode.DoesNotExist:
+            raise NotFound(detail="PythonNode not found.")
+
+        successful_sessions_exist = Session.objects.filter(
```
VitaliiPovoroznykHYS: We don't really need this query, later on we will find out that there is something or None. You basically need that only for this message ""detail": "No successful sessions found for this flow.", but it's okay if we don't have it, frontend does not care

src/django_app/tables/views/views.py
```diff
@@ -1020,3 +1024,97 @@ def post(self, request):
...
+        python_node_name = f"{python_node.node_name} #{python_node.pk}"
+        message = (
+            GraphSessionMessage.objects.filter(
+                session__graph_id=python_node.graph_id,
+                session__status=Session.SessionStatus.END,
+                name=python_node_name,
+                message_data__message_type="start",
+            )
+            .order_by("-session__created_at", "-created_at")
+            .first()
```
VitaliiPovoroznykHYS, line 1010: before .first() add this `.values_list("message_data__input", flat=True)`, this will make this query much faster, but you will need to modify your later code, cause previously a missing input key would return {} as the default, but now you'll get None for both "no row found" and "row found but no input key"

VitaliiPovoroznykHYS (changes requested):

VitaliiPovoroznykHYS (commented):
So far backend looks good, nothing to change

EdwardPaseniukHYS (approved):

VitaliiPovoroznykHYS (approved):

PR #452, opened 2026-04-15
chore(EST-1701-nodes-bulk-save-endpoint): implement nodes bulk save endpoint saving

EdwardPaseniukHYS:
- implemented nodes bulk save logic with temp id field handling for nodes and connections
- Refactored visual-programming graph load/save flow into modular utils/load/* and utils/save/* pipelines, replacing legacy graph service implementations.
- Aligned connection UX with validation rules in the flow graph, including Decision Table port behavior (drag/reassign eligibility, dynamic decision roles, and highlight consistency).


ViraKalenykHYS (approved):
Looks great, approved!

PR #459, opened 2026-04-20
Feature EST 2487: View version history

PavloNesterukHYS:
Add functionality that helps users to list all active versions of flow:

- When a user clicks "Version History" in the flow editor header, a panel/dialog opens showing all versions for the current flow
- When versions exist, each entry displays: name, date saved, and description visible on hover
- When no versions exist, the panel shows an empty state message (e.g., "No versions saved yet")
- When the list has many versions, it is scrollable
- When the user edits a version's name or description, the change is saved immediately
- When the user closes the panel/dialog, they return to the editor as normal


src/django_app/tables/views/model_view_sets.py
```diff
@@ -862,6 +867,30 @@ def save_flow(self, request, pk=None):
         refreshed = self.get_queryset().get(pk=pk)
         return Response(GraphSerializer(refreshed).data, status=status.HTTP_200_OK)
 
+    @action(detail=True, methods=["post"], url_path="save-version")
```
VitaliiPovoroznykHYS: I would recommend to use separate path for versions. Like /graph-versions/ not graphs/{id}/save-version/

VitaliiPovoroznykHYS (changes requested):

src/django_app/tables/views/model_view_sets.py
```diff
@@ -879,6 +886,56 @@ def get_queryset(self):
         )
 
 
+class GraphVersionViewSet(viewsets.ModelViewSet):
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = ["graph_id"]
+
+    def get_queryset(self):
+        qs = GraphVersion.objects.filter(is_active=True, deleted_at__isnull=True)
```
VitaliiPovoroznykHYS: This would be better if we have that on the model side, cause right now anytime we want to query anything we need to include filter to exclude is_active.

Much better way to do that is to create your own custom manager for django model. Something like this:

```from django.db import models

class ActiveManager(models.Manager):
    def get_queryset(self):
        return super().get_queryset().filter(is_active=True)


class MyModel(models.Model):
    name = models.CharField(max_length=255)
    is_active = models.BooleanField(default=True)

    objects = ActiveManager()        # default manager (filtered)
    all_objects = models.Manager()   # unfiltered access
```

In this way we don't need to think all the time about active objects and we can reuse that manager for other models that have soft delete. We also don't need to touch all the other code in the app

VitaliiPovoroznykHYS (changes requested):

src/django_app/tables/models/base_models.py
```diff
@@ -128,6 +128,20 @@ class Meta:
         abstract = True
 
 
+class SoftDeleteMixin(models.Model):
+    deleted_at = models.DateTimeField(
+        db_index=True,
```
VitaliiPovoroznykHYS: remove db_index=True for now, we will see do we need it later on or no

src/django_app/tables/views/model_view_sets.py
```diff
@@ -879,6 +887,52 @@ def get_queryset(self):
...
+    def get_queryset(self):
+        manager = (
+            GraphVersion.all_objects if self.action == "all" else GraphVersion.objects
+        )
+        qs = manager.all()
+        if self.action in ("list", "all"):
+            qs = qs.defer("snapshot", "dependencies")
+        return qs
+
+    def get_permissions(self):
```
VitaliiPovoroznykHYS: let's remove permissions for now, at this point of time it's okay if everyone can see what was deleted

VitaliiPovoroznykHYS (changes requested):

KarolinaKudrytskaHYS (approved):

src/django_app/tables/views/model_view_sets.py
```diff
@@ -897,6 +909,10 @@ def create(self, request, *args, **kwargs):
             status=status.HTTP_201_CREATED,
         )
 
+    @action(detail=False, methods=["get"], url_path="all")
+    def all(self, request, *args, **kwargs):
```
VitaliiPovoroznykHYS, line 910: Okay, but make sure frontend does not use this endpoint cause it will show all version (deleted as well)

VitaliiPovoroznykHYS (approved):

PR #472, opened 2026-04-23
feat(EST-2067-FE): implement nodes routing in flow editor

ViraKalenykHYS:
### Scope:
Add support for nodes routing in the flow editor, including waypoint-based editing of existing connections and consistent rendering/interaction behavior.

### Implemented work:

**1) Smart routing and loop visibility**

- Added detection and dedicated rendering for backward connections so they are visually distinguishable from regular forward connections.
- Backward routes are displayed as dashed connections and routed in a more visible way instead of blending into regular links.
- Added automatic rerouting for non-manual connections when a connection path overlaps node areas, so the line rebuilds itself to avoid passing through nodes where possible.
- This auto-rerouting applies only to automatically calculated routes and should not override user-defined manual routing.

**2) Manual connection routing**

- Added support for waypoint-based editing of existing connections in the flow editor.
- Users can manually adjust a connection path by dragging waypoints and shaping the route visually.
- Manual routing should persist for edited connections and be used instead of automatic recalculation unless the connection is reset/rebuilt by system logic.

**3) Gradient connection highlighting**

- Added support for gradient-colored connection rendering in the flow editor.
- This provides clearer visual distinction and improves readability of connections in the graph.
- Gradient rendering should work consistently with the updated routing behavior.


ElenaPrygunHYS (approved):

ElenaPrygunHYS:
Ready for merge

PR #475, opened 2026-04-27
Feature EST-2544: Restore a version

PavloNesterukHYS:
### Functionality
- When a user clicks "Restore" on a version, a warning dialog appears with Save & Restore / Just Restore / Cancel options
- When the user selects "Save & Restore," a version named "Before restore to [version name]" is created from the current state, then the selected version is restored
- When the user selects "Just Restore," the selected version is restored immediately without saving the current state
- When the user selects "Cancel," nothing happens and they return to the version history
- When a version is restored, the flow canvas reloads completely to reflect the restored state
- When a version is restored successfully, an informative toast is shown
- When a restored version references deleted tools/agents, the flow restores and a warning indicator (red button/icon) appears on the canvas
- When the user clicks the warning indicator, a list of all missing dependencies is shown with details about which nodes are affected

### Back-end
- Actual restore feature almost inherit from and based on import/export system: 
    - Collects all dependencies and make snapshot using `GraphStrategy` export
    - Using "wipe-and-recreate" method based on import  
- Available compatibility for older versions, based on converting system of import/export. *When adding new converter always keep in mind it could be used for both import flow and restore version.*
- Functionality is covered in unit test for:
    - `handlers/` strategies
    - manager layer
    - service layer
    - api calls


src/django_app/tables/graph_versioning/strategy.py
```diff
@@ -30,3 +58,336 @@ def collect_dependencies(self, graph: Graph) -> dict:
...
+                "python_code_id", flat=True
+            )
+        )
+        python_code_ids.update(
+            WebhookTriggerNode.objects.filter(graph=graph).values_list(
+                "python_code_id", flat=True
+            )
+        )
+
+        for relation_name in (
```
VitaliiPovoroznykHYS: move this list of relations to `constants.py`

src/django_app/tables/graph_versioning/strategy.py
```diff
@@ -30,3 +58,336 @@ def collect_dependencies(self, graph: Graph) -> dict:
...
+            orphan_ids = python_code_ids - shared_ids
+            if orphan_ids:
+                PythonCode.objects.filter(id__in=orphan_ids).delete()
+
+    def _update_graph_scalars(self, graph: Graph, snapshot: dict) -> None:
+        """
+        Updates graphs fields from version snapshot
+        """
+        update_fields = []
+        for field in _GRAPH_SCALAR_FIELDS:
```
VitaliiPovoroznykHYS: I don't really like that we have to list scalar fields manualy. They should be detected automatically, it will be very annoying to work with. Whenever we add a new fields to the graph model, we will also need to add them to that scalar fields list.

src/django_app/tables/graph_versioning/strategy.py
```diff
@@ -1,5 +1,33 @@
...
+class _MissingSets:
+    """Dataclass that holds all missing deps"""
+
+    crews: set
+    subgraphs: set
+    llm_configs: set
+    webhooks: set
 
 
 class GraphVersioningStrategy:
```
VitaliiPovoroznykHYS: So we are using this versioning strategy, but we have an issue here. What if graph was saved long time ago? We added new fields, new models, etc. In current scenario we cannot really restore something like that. Also since we are using logic of export here, we really should use logic of import here as well. At least we should run our snapshot through import/export migrations before restoring a version. (backward compatibility issue)

VitaliiPovoroznykHYS (changes requested):
There are some changes that needs to be done on the backend side

VitaliiPovoroznykHYS (changes requested):
Backend requires some changes, mostly related to backward compatibilirt

src/django_app/tables/graph_versioning/manager.py
```diff
@@ -30,3 +61,363 @@ def collect_dependencies(self, graph: Graph) -> dict:
...
+    def _filter_nodes(
+        self, nodes: list[dict], missing_sets: _MissingSets
+    ) -> tuple[list[dict], set[int], list[dict]]:
+        """Checks all graph nodes that rely on dependencies and skip them"""
+
+        kept_nodes: list[dict] = []
+        skipped_node_ids: set[int] = set()
+        warnings: list[dict] = []
+
+        for node in nodes:
```
VitaliiPovoroznykHYS, line 110: Use Strategy pattern for this

src/django_app/tables/models/graph_models.py
```diff
@@ -88,7 +88,11 @@ class CrewNode(BaseNode):
     graph = models.ForeignKey(
         "Graph", on_delete=models.CASCADE, related_name="crew_node_list"
     )
-    crew = models.ForeignKey("Crew", on_delete=models.CASCADE)
+    crew = models.ForeignKey(
```
VitaliiPovoroznykHYS: Don't set crew to NULL, cause in future we will have Crew only as part of current flow, it will never be separated

src/django_app/tables/models/graph_models.py
```diff
@@ -180,7 +184,10 @@ class SubGraphNode(BaseNode):
         "Graph", on_delete=models.CASCADE, related_name="subgraph_node_list"
     )
     subgraph = models.ForeignKey(
-        "Graph", on_delete=models.CASCADE, related_name="as_subgraph"
+        "Graph",
+        on_delete=models.SET_NULL,
+        related_name="as_subgraph",
+        null=True,
     )
     # TODO: maybe SET_NULL on delete?
```
VitaliiPovoroznykHYS: Nice, but now remove this TODO :)

src/django_app/tables/apps.py
```diff
@@ -18,6 +18,7 @@ def ready(self):
         import tables.signals.naive_rag_signals
         import tables.signals.webhook_signals
         import tables.import_export.version_conversions.convertions
+        from tables.import_export.version_conversions.base import VersionConverter
```
VitaliiPovoroznykHYS: Move this to line 30, so we have similar imports grouped

VitaliiPovoroznykHYS (changes requested):
Some changes requested

VitaliiPovoroznykHYS (changes requested):
Fix current issues

VitaliiPovoroznykHYS:
Also add unit tests to cover this functionality

frontend/src/app/features/flows/components/version-history-panel/version-history-panel.component.scss
```diff
@@ -70,10 +70,17 @@
...
     box-shadow: 0 2px 8px rgba(0, 0, 0, 0.3);
+    border: 2px solid transparent;
+    cursor: pointer;
 
     &:hover {
         background-color: rgba(255, 255, 255, 0.07);
     }
+
+    &--selected {
+        border-color: #685fff;
```
AndreyKrasnoyarskiy: May be better to avoid hardcodded colors

frontend/src/app/features/flows/components/restore-warnings-dialog/restore-warnings-dialog.component.html
```diff
@@ -0,0 +1,73 @@
...
+        <app-button
+            type="ghost"
+            (click)="dialogRef.close()"
+            >Close</app-button
+        >
+        <app-button
+            type="primary"
+            (click)="dialogRef.close()"
+            >Got it</app-button
+        >
```
AndreyKrasnoyarskiy: Why the same action on different buttons?

frontend/src/app/features/flows/components/version-history-panel/version-history-panel.component.ts
```diff
@@ -100,7 +118,7 @@ export class VersionHistoryPanelComponent implements OnInit {
         const value = this.editingValue.trim();
         const originalValue = field === 'name' ? version.name : version.description || '';
 
-        if (!value || value === originalValue) {
+        if ((field === 'name' && !value) || value === originalValue) {
```
AndreyKrasnoyarskiy, line 129: Better not to use hardcodded words

AndreyKrasnoyarskiy (approved):

src/django_app/tables/graph_versioning/manager.py
```diff
@@ -0,0 +1,345 @@
...
+        id_mapper = IDMapper()
+        for entity_type_value, ids in available_deps.items():
+            entity_type = _DEPENDENCY_ENTITY_TYPES.get(entity_type_value)
+            if entity_type is None:
+                continue
+            for entity_id in ids:
+                id_mapper.map(entity_type, entity_id, entity_id, was_created=False)
+        return id_mapper
+
+    def convert_snapshot_to_current_version(self, snapshot: dict) -> dict:
```
VitaliiPovoroznykHYS, line 330: convert_snapshot_to_current_version builds a fake bundle containing only graph data ({GRAPH: [...], version, main_entity}) and passes it to VersionConverter.convert, which is designed for full import bundles that include every entity type (crews, agents, LLMs, etc.).

This works today only because IMPORT_VERSION = 1 and no converters exist yet. The moment we bump the version and add a real converter, that converter will assume the bundle has all entity shapes and either silently no-op or raise a KeyError — and restore will break for every old snapshot, with no obvious link back to this code.

Suggested fix:

add a clear contract on VersionConverter.convert stating that partial bundles are allowed, and require all future converters to guard against missing keys

src/django_app/tables/graph_versioning/constants.py
```diff
@@ -0,0 +1,46 @@
+from tables.import_export.enums import EntityType
+from tables.models import (
+    Crew,
+    Graph,
+    LLMConfig,
+    WebhookTrigger,
+)
+
+_EXCLUDED_GRAPH_SCALARS = (
```
VitaliiPovoroznykHYS, line 9: add "uuid" to _EXCLUDED_GRAPH_SCALARS alongside id

src/django_app/tables/graph_versioning/services.py
```diff
@@ -23,3 +30,37 @@ def save_version(
             snapshot=snapshot,
             dependencies=light_deps,
         )
+
+    @transaction.atomic
+    def restore_version(self, version: GraphVersion, *, backup: bool = False) -> dict:
```
VitaliiPovoroznykHYS, line 35: The auto-backup GraphVersion is created inside the same @transaction.atomic block that wipes and recreates the graph. If anything in the restore raises, the whole transaction rolls back — including the backup row — but the method still returns auto_backup_version_id to the caller. On failure that ID points at a row that no longer exists

Make it explicit (docstring + return contract) that auto_backup_version_id is only meaningful when the call succeeds, and don't return it on the failure path

VitaliiPovoroznykHYS (changes requested):
There are some changes that needs to be done

VitaliiPovoroznykHYS (changes requested):
Should be last set of changes :)

VitaliiPovoroznykHYS (approved):

PR #477, opened 2026-04-27
Feature/est 1529 tools configuration

MaksymYevstratievHYS:


BogdanTimchenkoHYS:
## What

Rebuilt args_schema generation for Python code tools: extracted it from the inline tool factory into a dedicated `tool_factories` package, and made it Pydantic-driven via a shared `Variable` model — replaces a tangled in-line builder that mixed schema, defaults, and type conversion in one class. Completed support for nested `object` / `array` variables. As a side-fix, replaced obsolete `"type": "integer"` with `"number"` across built-in tools to unblock the UI.

## Why

The previous `ArgsSchemaFactory` mixed schema construction, default casting, and type mapping into one hard-to-read class. With object/array on the roadmap this would have grown unmaintainable. The new version reads validated Pydantic Variables directly and is short enough to skim in one pass. The shared model also unifies the type contract between Django, crew, and any future service.

## How

- `src/shared/models/variables.py` — discriminated-union `Variable` / `NestedVariable` (5 types). Each variant declares its own `python_type`.
- `src/shared/dotdict/` — `DotDict` / `DotList` expose the Pydantic protocol via lazy imports; user `main()` gets attribute-style access to object/array values.
- `src/crew/services/crew/tool_factories/` — new package: `ArgsSchemaFactory`, `CrewToolDynamicFactory`, `PythonCodeToolFactory` (replaces removed `proxy_tool_factory`).
- `docs/tools/` and built-in `args_schema.json` files updated to match the new schema.

## Side Effects

- Object / Array variables now arrive in `main()` as `DotDict` / `DotList` (subclasses of dict/list). `isinstance(value, dict)` still passes; strict `type(value) is dict` would not — corner case worth flagging.
- Wire format change in this PR (`items` → `item`, no `obj`/`list` aliases) — FE on the same branch is aligned; do not merge BE alone.
- Built-in tools re-upload on `upload_tools` and end up with `number` instead of `integer` — verify FE table view handles the rename.

## Related

- [EST-1529 — Tools configuration (umbrella)](https://hys-enterprise.atlassian.net/browse/EST-1529)
- [EST-2807 — Agent input ignores default value | BE](https://hys-enterprise.atlassian.net/browse/EST-2807)
- [EST-2808 — JSON/table display broken for old tools | BE](https://hys-enterprise.atlassian.net/browse/EST-2808)

src/crew/main.py
```diff
@@ -1,6 +1,10 @@
 import asyncio
 import os
 import sys
+from pathlib import Path
+
+# Add "src/" to system path to import shared module
+sys.path.insert(0, str(Path(__file__).parent.parent))
```
ArturDrohomeretskyiHYS: # Imports: align with the repo-wide convention (Django / realtime / knowledge)

The move from `src.crew.*` to bare internal imports is the right direction — keep that. But the shared-module change and the `sys.path` insert break consistency with the rest of the codebase. Here's the standard and what to change.

## The convention every other service follows

- **Internal modules → bare, rooted at the service dir.** Django uses `from tables...`, so crew should use `from services...`, `from models...`, `from utils...`.  Your internal-import change already does this.
- **Shared → always `from src.shared...`.** `shared/` lives *outside* the service dir (it's a sibling under `src/`), so it's addressed through the common `src.` namespace. This is anchored on `PYTHONPATH=/app` in every Dockerfile, and on `PYTHONPATH=${workspaceFolder}` in the launch configs.

## 1. Revert shared imports to keep the `src.` prefix

```python
# change this back:
from shared.models import PythonCodeData
# to (matches django / realtime / knowledge):
from src.shared.models import PythonCodeData
```

## 2. Remove the `sys.path.insert` hack from `main.py`

```python
# delete these lines:
from pathlib import Path
sys.path.insert(0, str(Path(__file__).parent.parent))
```

Reasons it shouldn't land:

- It only exists to prop up the bare `shared.` style — unneeded once shared is `src.shared` (the `/app` root is already on `PYTHONPATH`, so `src.shared` resolves with zero code).
- It's an import-time side effect that must run *before* any `from shared...` import. isort / ruff / black will reorder imports above it and silently break resolution.
- It only covers `main.py`. pytest, and any other entrypoint that imports a module doing `from shared...`, won't have the path injected → import errors. (`src.shared` works everywhere because it rides on `PYTHONPATH`, not on one file.)
- Mixing bare `shared.` (this PR) with `src.shared.` (rest of repo) loads `shared.models` under **two module identities** — duplicate class objects, broken `isinstance`, duplicated module-level state.

Keep `import sys` (it's still used for `sys.argv`), only drop the `Path` import and the insert.

## 3. Resulting target for the example block

```python
from models.state import State
from services.graph.events import StopEvent
from services.graph.exceptions import StopSession
from services.graph.nodes import BaseNode
from services.run_python_code_service import RunPythonCodeService
from src.shared.models import PythonCodeData   # ← src. stays only on shared
```

## 4. Make it run locally the same way knowledge / django already do

No code hack, just the path the rest of the repo uses.

- **VSCode** — the `Crew` config in `.vscode/launch.json` is the only one missing `PYTHONPATH`. Add it to match the `Knowledge` config:

  ```jsonc
  {
    "name": "Crew",
    "type": "debugpy",
    "request": "launch",
    "program": "${workspaceFolder}\\src\\crew\\main.py",
    "cwd": "${workspaceFolder}",
    "envFile": "${workspaceFolder}\\src\\debug.env",
    "env": { "PYTHONPATH": "${workspaceFolder}" }
  }
  ```

  (`program` dir `src/crew` → `services.*`; `PYTHONPATH=${workspaceFolder}` → `src.shared.*`. Same two roots Docker has.)

- **Terminal** (from repo root):

  ```powershell
  $env:PYTHONPATH="$PWD"; python src\crew\main.py
  ```

## 5. pytest

Tests import `src.shared`, so the repo root must be on the path too. Update `crew/pytest.ini`:

```ini
pythonpath = . ../..
```

(`.` = crew dir for `services.*`; `../..` = repo root for `src.shared.*`.)

## 6. Consistency check before merge

Apply the bare-internal rule uniformly. Don't leave any `from src.crew.*` imports mixed with the new `from services.*` ones in the files you touched — the mix reintroduces the same double-module problem described above. Ideally we convert the remaining `src.crew.*` imports repo-wide in a follow-up so the whole service is one style.

---

**Net:** internal = bare (`services.*`, `models.*`), shared = `src.shared.*`, no `sys.path` mutation, no Dockerfile change. Works in Docker (unchanged) and locally via the launch config — identical to how django and knowledge already behave.

ArturDrohomeretskyiHYS (changes requested):

src/crew/uv.lock
ArturDrohomeretskyiHYS, line 1: add to .gitignore

ArturDrohomeretskyiHYS (approved):

KarolinaKudrytskaHYS (approved):

PR #479, opened 2026-04-27
Feature/est 2418 password recovery

ArturDrohomeretskyiHYS:


frontend/src/app/features/auth/components/forgot-pass-page/forgot-password-page.component.ts
```diff
@@ -0,0 +1,62 @@
...
+        validators: [Validators.required, Validators.email],
+    });
+
+    onRequestReset(): void {
+        this.emailControl.markAsTouched();
+        if (this.emailControl.invalid) return;
+        const email = this.emailControl.getRawValue().toString();
+
+        this.loading.set(true);
+        this.authService
```
EdwardPaseniukHYS, line 53: here you can use TAP from rxjs for   this.submittedEmail.set(this.emailControl.getRawValue());
and finalize() for     this.loading.set(false);
                this.state.set('email-sent');

i believe in this code in case of error in the request the logic in subscribe will not run

frontend/src/app/features/auth/components/reset-password-page/reset-password-page.component.ts
```diff
@@ -0,0 +1,112 @@
...
+        const data = {
+            token: this.token,
+            new_password: this.password.toString(),
+        };
+
+        this.loading.set(true);
+        this.authService
+            .confirmResetPassword(data)
+            .pipe(takeUntilDestroyed(this.destroyRef))
+            .subscribe({
```
EdwardPaseniukHYS, line 82: here the same, from what i know better practise to use tap, finalize and catcherror . you repeat loading set false twice , violation of DRY. but having that this code is working its up to you whether to change or not

EdwardPaseniukHYS (approved):

VitaliiPovoroznykHYS (approved):

PR #486, opened 2026-04-29
feat(EST-2660): auto-update graph.updated_at on node changes

NikitaZalohaHYS:


VitaliiPovoroznykHYS (changes requested):
We want to change updated_at when any graph entity was changed. In current approach it's being changed only when nodes changed, but we also have Edges that are not BaseGlobalNode

VitaliiPovoroznykHYS (changes requested):
Current approach works, but we don't change updated_at when Edges updated/changed, since they are not inherited from BaseGlobalNode

VitaliiPovoroznykHYS (approved):

PR #489, opened 2026-04-29
fix(EST-847-FE): Repeatitive requests on project page

AndreyKrasnoyarskiy:


BohdanSydirHYS (approved):
Consider moving storage services from 'features/configure-models/services/llms' to 'shared/services'

AndreyKrasnoyarskiy:
> Consider moving storage services from 'features/configure-models/services/llms' to 'shared/services'

Very good point, will do it

PR #493, opened 2026-04-30
fix(EST-2742): fix integration tests

NikitaZalohaHYS:


VitaliiPovoroznykHYS (changes requested):
Everything related to tools in manager container is deprecated, therefore no need to do any fixes for them specifically, we should just remove the tests

src/manager/services/tool_image_service.py
```diff
@@ -13,10 +14,18 @@
 
 class ToolImageService:
```
VitaliiPovoroznykHYS: We don't need this service at all, everything that related to Tools in separate docker containers is no longer used

Unless I am missing something

integration_tests/utils/utils.py
```diff
@@ -441,28 +514,53 @@ def create_crew_node(
...
-        f"{DJANGO_URL}/crewnodes/", json=crew_node_data, headers={"Host": rhost}
+        f"{DJANGO_URL}/crewnodes/", json=crew_node_data, headers=get_headers()
     )
     validate_response(response)
     return response.json()["id"]
 
 
-def create_edge(start_key: str, end_key: str, graph: int) -> int:
-    edge_data = {"start_key": start_key, "end_key": end_key, "graph": graph}
+def create_llm_node(
```
VitaliiPovoroznykHYS, line 523: Also remove all tests/code related to LLM node, it is no longer used as well

src/manager/tools_config.json
VitaliiPovoroznykHYS, line 1: I think we don't need crewai tools as well, all our tools are just custom python tools now, unless I missed something here as well

VitaliiPovoroznykHYS (changes requested):

VitaliiPovoroznykHYS (approved):
`src/tool/services/__init__.py` + `base_models.py` are pure re-exports, nothing imports them. Just remove them.

PR #504, opened 2026-05-07
Fix/memory issues

VitaliiPovoroznykHYS:
Fixed memory issues by closing pubsub connection after closing SSE stream. Also added few options for better memory management on the server. Also added background job to trim memory allocation in django app once in a minute. There are also some bug fixes related to gpt5 and tool calling


PavloNesterukHYS (approved):
Super, nice work)

PR #507, opened 2026-05-11
Feature/est 2420 user management

ArturDrohomeretskyiHYS:
## Story 5 (EST-2420) — User Management (Superadmin and Org Admin)

Adds the admin-panel user-management surface on top of the Story 4 RBAC
foundation. 9 endpoints under `/api/admin/users/...` and
`/api/admin/organizations/{org_id}/users|assign-users/...`.
Authorization splits two ways: superadmin-only for cross-org operations,
superadmin-or-Org-Admin for per-org membership operations.

## New classes

- **`UserManagementService`** — single orchestrator for every Story 5
  read/write. Views are thin wrappers that validate input and delegate.
  Every write is wrapped in `transaction.atomic()` and locks the
  contested row with `SELECT FOR UPDATE` before any guard runs, which
  is what makes the last-superadmin / last-Org-Admin checks race-safe.
  Translates `IntegrityError` into typed domain exceptions.
- **`UserManagementGuards`** — four pure invariant checks
  (last-active-superadmin, last-Org-Admin single-row, last-Org-Admin
  net-effect across a batch, role-is-assignable). Static methods so
  they can be called without instance state; expected to run inside
  the service's transaction *after* the contested rows are locked.
  Reserved for system-integrity invariants only — caller-relationship
  UX rules live inline in the service method that owns them (see
  Two-tier guard rule below).
- **`UserValidationService(BaseRBACValidator)`** — request-shape
  validation for the new endpoints. Aggregates per-field errors and
  raises a single `FormValidationError` carrying the structured
  `errors[]` list. Redacts `password` in echoed values.
- **`IsSuperadminOrOrgAdmin`** permission class — accepts
  `is_superadmin=True` globally, OR membership in the URL-provided
  organization with the built-in `Org Admin` role. Defense-in-depth:
  the service re-checks the same membership inside its transaction.
- **`UserAdminViewSet`** — cross-org user management. Superadmin only.
  Paginated list, create-with-optional-initial-membership,
  grant/revoke superadmin.
- **`OrganizationMembershipAdminViewSet`** — per-org membership
  management nested under `/admin/organizations/{org_id}/`. Allows
  superadmin globally or Org Admin of the URL org. Surfaces:
  list / create-and-link a brand-new user / change role / remove,
  plus the new batch `assign-users` action.
- **7 new domain exceptions** in `rbac_exceptions.py`
  (`EmailAlreadyExistsError`, `MembershipAlreadyExistsError`,
  `LastSuperadminError`, `LastOrgAdminError`,
  `InvalidRoleAssignmentError`, `RoleNotFoundError`,
  `CannotSelfAssignError`) — all extend `CustomAPIExeption` so the
  project envelope renders them automatically.

## Model change

`OrganizationUser.unique_together = (("user", "org"),)` re-expressed as
a named `UniqueConstraint` (`rbac_organization_user_user_org_uniq`).
Behavior identical; the constraint name is now stable and grep-able so
the service can catch `IntegrityError` and translate to
`MembershipAlreadyExistsError`.

## Legacy cleanup

The plain `OrganizationUserViewSet` at `/api/organization-users/` was
retargeted in Story 0 onto the new RBAC `OrganizationUser` table but
never received permission checks — any authenticated user could PATCH
themselves into the Org Admin role. Removed entirely (`ViewSet`,
`ModelSerializer`, URL registration). Story 5 endpoints are now the
only surface for `OrganizationUser` CRUD.

## Batch upsert assign-users + single-mode add-member

The `POST /admin/organizations/{org_id}/users/` endpoint is now
**create-and-link only**. The previous link-existing-user mode (Mode A,
`{"user_id": ..., "role_id": ...}`) was removed — both creating and
reassigning existing users now go through a new dedicated batch
endpoint:

`POST /admin/organizations/{org_id}/assign-users/` is an **upsert**.
For each `{user_id, role_id}` row: if no membership exists, it's
created; if one exists with a different role, the role is updated; if
the role is unchanged, the row is a no-op (still returned). Up to 100
rows per request. All-or-nothing inside one transaction. Returns 200
OK with `{"created": [...], "updated": [...]}` preserving submission
order in each array.

Two endpoint-specific rules:
- **Self-assignment**: a non-superadmin caller cannot include their own
  `user_id` in the batch (`cannot_self_assign`). Superadmin bypasses.
  The single-row `PATCH` endpoint exists for deliberate self-changes.
- **Net-effect last-Org-Admin guard**: the batch is rejected
  (`last_org_admin`) if applying it would leave the org with zero
  Org Admins. Demote-one-promote-another in the same batch is fine.

This is a breaking API change with no compat shim.

## Two-tier guard rule

Story 5 distinguishes three tiers of access checks, documented in the
progress log:

1. **Permission gate** (`IsSuperadmin*` classes) — superadmin always bypasses.
2. **System-integrity invariant** (`UserManagementGuards`) — applies to
   *all* callers including superadmin (last-superadmin, last-Org-Admin,
   role-assignability).
3. **Caller-relationship UX rule** (inline in the service method,
   not in `UserManagementGuards`) — superadmin bypasses. The self-assign
   rule on `assign_users` is the first instance.

Future stories add new caller-aware UX rules inline; `UserManagementGuards`
stays reserved for system-integrity invariants.

## Hardening

URL detail routes on `UserAdminViewSet` and `OrganizationAdminViewSet`
previously called `int(pk)` with no URL-level type guard, so a string
pk surfaced `ValueError` as HTTP 500. Both ViewSets now constrain the
pk regex to digits, so non-numeric segments return 404 from the URL
resolver before view code runs.

## Test coverage

Permission gates (anonymous, Member, Org Admin, API-key bound),
happy paths for all 9 endpoints, batch upsert behavior (reassign
existing role, no-op on same role, mixed create-and-reassign in one
batch), net-effect Org Admin guard (block last-admin demote, allow
demote-with-promote in same batch), edge cases (email conflict,
invalid role assignment, last-active-superadmin, last-Org-Admin on
single-row delete and demote, zero-orgs steady state), validation
errors (missing fields, weak password redacted, malformed email,
unknown ids, batch over-cap, duplicate user_id in batch, non-int
types), self-assign rule for both branches (Org Admin blocked,
superadmin allowed), and regression coverage for the pk-500 fix.


ArturDrohomeretskyiHYS:
merged here: https://github.com/EpicStaff/EpicStaff/pull/526

PR #518, opened 2026-05-13
Export sessions

ViraKalenykHYS:


src/django_app/tables/import_export/strategies/session.py
```diff
@@ -0,0 +1,54 @@
...
+    @staticmethod
+    def csv_row_mapper(m: dict) -> dict:
+        return {
+            "id": m["id"],
+            "session_id": m["session_id"],
+            "created_at": m["created_at"],
+            "name": m["name"],
+            "execution_order": m["execution_order"],
+            "message_type": (m.get("message_data") or {}).get("message_type", ""),
+            "message_data": json.dumps(m.get("message_data")),
```
VitaliiPovoroznykHYS: We don't want to have JSON fields in CSVs, since it's not filtrable for someone using it externally as audit. Think of a way how can we represent this kind of potentially nested dicts in CSV

VitaliiPovoroznykHYS (commented):
Want to flag a couple things, but it's not required to change this right now, we can leave it as technical debt

**1. CSV_FIELDS / csv_row_mapper shouldn't live on the export strategy base.**

It's format-specific knowledge on a class that shouldn't care about formats. Add XLSX next and we're adding XLSX_FIELDS, XLSX_ROW_MAPPER, etc. — base class grows with every format. Also the attrs are Optional but `SessionViewSet.__init__` treats them as required, so a new entity without them silently breaks CSV instead of erroring.

**2. JSON and CSV aren't really the same kind of thing.**

JSON is bidirectional — it's our canonical export and the import input. CSV (and any future XLSX/etc) is a one-way view over that export. Modelling them as sibling strategies in one dict is what pulled CSV-specific stuff onto the base in the first place.

Also: import_export_service.py falls back to JSON when format isn't recognized. Should 400 — otherwise user asks for CSV, gets JSON, no clue why.

**What I'd do instead — split it into 3 layers:**

1. **Export strategies** (`strategies/`, what we have today) — only handle JSON round-trip: entity → dict, dict → entity. Remove the CSV stuff from here.

2. **Tabular projections** (new folder, e.g. `tabular/`) — one small class per entity that knows how to flatten that entity's exported dict into rows for spreadsheet-style formats. Holds the `fields` list and a `project(exported_dict) -> rows` method. Only entities where "flat rows" makes sense get one (Session yes, Graph probably no).

3. **Renderers** (`export_format_strategies.py`) — pure format code, no entity knowledge.
   - `JsonFormatStrategy`: takes a dict, writes JSON.
   - `CsvFormatStrategy`: takes a `TabularProjection` in `__init__`, uses it to turn the dict into rows, writes CSV. Doesn't import `Session` or any entity.
   - Future `XlsxFormatStrategy`: same deal, takes any `TabularProjection`.

**How the pieces connect at request time:**

- JSON request: `Session export strategy → dict → JsonFormatStrategy → response`
- CSV request: `Session export strategy → dict → SessionTabularProjection → rows → CsvFormatStrategy → response`

Same export dict feeds both paths. CSV path just has the projection step in the middle.

**Why this is better:**

- Adding XLSX = write one `XlsxFormatStrategy`. Done. Every entity that already has a projection works with it automatically.
- Adding a new entity that needs CSV/XLSX = write one projection. Done. Every tabular renderer works with it automatically.
- Base export strategy never grows when we add formats.
- An entity without a projection can't be CSV-exported — and that's enforceable with a clear 400, not a silent fallback.

ElenaPrygunHYS (approved):

VitaliiPovoroznykHYS (approved):

PR #523, opened 2026-05-14
Feature/EST-1429 add drift search and EST-1428 add global search

VasylCherkesHYS:
[FRONTEND_INTEGRATION_GUIDE.md](https://github.com/user-attachments/files/28096020/FRONTEND_INTEGRATION_GUIDE.md)


VasylCherkesHYS:

src/crew/libraries/crewAI/src/crewai/agents/crew_agent_executor.py
```diff
@@ -33,8 +33,22 @@ class ToolResult:
...
+    "If that information says the documents do not cover the question, or does not "
+    "contain what was asked, say so plainly — do not invent or guess an answer.\n"
+)
+EMPTY_KNOWLEDGE_KEYWORD = (
+    "\n\nNO EXTRACTED KNOWLEDGES!\n"
+    "\nGrounding rule: no relevant information was found in the knowledge base for "
+    "this question. If it asks for factual information that would be expected to "
+    "come from the knowledge base, state that the available documents do not cover "
+    "it — do not answer it from your own general or training knowledge.\n"
+)
```
VasylCherkesHYS: here we rewrote the built-in CrewAI agent engine (the loop "think → use tool → answer"). We added our own grounding rules into the text that is put around the found knowledge (RAG): the markers RELEVANT KNOWLEDGES / END OF RELEVANT KNOWLEDGES / NO EXTRACTED KNOWLEDGES. These rules forbid the agent to invent facts from its own training knowledge and make it answer only from the documents; and if there is nothing in the documents — to say honestly "not covered". It is used in the crew service when running an agent node with knowledge/RAG — to limit hallucinations.

src/crew/libraries/crewAI/src/crewai/agent.py
```diff
@@ -267,17 +265,13 @@ def execute_task(
...
-
-        if not agent_knowledge_snippet:
-            task_prompt += f"\n{EMPTY_KNOWLEDGE_KEYWORD}\n"
+            if agent_knowledge_snippet:
+                task_prompt += (
+                    f"{KNOWLEDGE_KEYWORD} \n\n{agent_knowledge_snippet}"
+                    f"\n{END_OF_KNOWLEDGE_KEYWORD}"
+                )
+            else:
+                task_prompt += f"\n{EMPTY_KNOWLEDGE_KEYWORD}\n"
```
VasylCherkesHYS: In execute_task() we added a RAG branch: if rag_type_id is set, the agent builds a knowledge query (ready-made or generated by the LLM at temp=0), searches the knowledge service over Redis, and puts the result into the prompt between the markers KNOWLEDGE_KEYWORD … END_OF_KNOWLEDGE (or EMPTY_KNOWLEDGE if nothing found). Why: so the agent answers strictly from the documents, not from its own knowledge. How it works now: before running the task, the agent pulls the relevant knowledge itself and gets it together with the grounding rules → fewer hallucinations.

src/django_app/tables/utils/llm_context_windows.py
VasylCherkesHYS, line 1: a single place that maps an LLM model name to its context window size. resolve_context_window() takes the value from litellm (knows ~500 models), else a user override, else a 16k fallback; and returns an is_trusted flag so the caller knows if it can relax token clamping. Why: to compute the token budget adaptively per model instead of hardcoding it. How it works: once a day one worker (via a file lock) fetches the fresh models JSON from GitHub and writes a shared snapshot; the other workers pick it up cheaply via an mtime stat — no extra network calls on the hot path.

src/knowledge/libraries/graphrag/graphrag/config/models/drift_search_config.py
```diff
@@ -57,6 +57,16 @@ class DRIFTSearchConfig(BaseModel):
...
+    relevance_threshold: float = Field(
+        description=(
+            "Minimum cosine similarity between the raw query and the most relevant "
+            "community report. Below this, DRIFT short-circuits to a canned "
+            "'documents do not cover this' answer instead of generating from "
+            "loosely-related context. Set per deployment (EpicStaff overlays it)."
+        ),
+        default=0.2,
+    )
+
```
VasylCherkesHYS: new field in the DRIFT search config if the similarity is below the threshold (default 0.2), DRIFT does not generate an answer but immediately returns a canned "documents do not cover this" - a relevance gate against hallucinations on off-topic queries

src/knowledge/libraries/graphrag/graphrag/prompts/query/drift_search_system_prompt.py
VasylCherkesHYS, line 1: Strengthened the grounding instructions in them: answer ONLY from the data in the tables/reports, never add general/training knowledge, and if the data does not cover the question — say plainly "the documents do not cover it" instead of making things up.

src/knowledge/libraries/graphrag/graphrag/query/structured_search/drift_search/search.py
```diff
@@ -210,11 +232,25 @@ async def search(
...
+                "DRIFT relevance %.4f (threshold %.4f)", relevance, threshold
+            )
+            if relevance < threshold:
+                logger.info(
+                    "Query not covered by indexed data — skipping generation."
+                )
+                return self._no_data_result(
+                    llm_calls, prompt_tokens, output_tokens, start_time
+                )
+
```
VasylCherkesHYS: added is the relevance gate: after build_context we take relevance (how close the query is to the top community report) and compare it with relevance_threshold; if it's lower, we immediately return the canned NO_DATA_ANSWER via _no_data_result() and do not run generation at all.

src/knowledge/rag/graph_rag/utils/search_prompt_builder.py
VasylCherkesHYS, line 1: Wraps the vendored prompts with our blocks: _DATA_GROUNDING_RULES (answer only from the data, not from general knowledge; if there's no data — say so plainly), _USER_PROMPT_WRAPPER (the user prompt goes last but cannot override the grounding rules), and for drift — _DRIFT_LENGTH_OVERRIDE (removes the hardcoded "exactly 2000 characters" so the intermediate answer isn't truncated before the reduce stage)

src/knowledge/rag/graph_rag/graph_rag_file_manager.py
VasylCherkesHYS, line 1: The path-resolution logic for the graph_data folder was moved into the shared src.shared.utils.graph_data_paths. Now the django_app container reuses it too — to reach output/*.parquet by the same paths and measure their size/metrics (for adaptive search-param suggestion).

src/knowledge/rag/graph_rag/graph_rag_strategy.py
```diff
@@ -357,6 +467,64 @@ def _load_parquet(self, root_folder: Path, filename: str) -> pd.DataFrame:
...
+
+        budget = self._context_token_budget(graphrag_config, search_method)
+        if not verify_grounded(query, str(response), context, graphrag_config, budget):
+            logger.warning(
+                f"Grounding guard rejected {search_method} answer as unsupported "
+                f"by retrieved context for query: [{query}]"
+            )
+            return ""
+
+        return response
```
VasylCherkesHYS: Method _apply_grounding_guard collapses the answer to an empty string in three cases: the answer is empty, it's the library's NO_DATA_ANSWER, or the LLM judge verify_grounded decides the answer isn't backed by the retrieved context. An empty string → _extract_chunks_from_context yields no chunks → the consumer treats the query as not covered by the knowledge base

src/knowledge/rag/graph_rag/graph_rag_strategy.py
```diff
@@ -259,6 +294,15 @@ def search(
...
 
+            # Step 4.5: Grounding guard — drop answers not backed by retrieved context
+            response = self._apply_grounding_guard(
+                query=query,
+                response=response,
+                context=context,
+                graphrag_config=graphrag_config,
+                search_method=search_method,
+            )
+
```
VasylCherkesHYS, line 290: final "guard" filter after any GraphRAG search (basic / local / global / drift) use for this method _apply_grounding_guard.

PR #529, opened 2026-05-18
Demo/feature/mod v2 flow assistant

igorpolishchukhys:


AndreyKrasnoyarskiy (approved):

src/django_app/tables/models/flow_assistant_models.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+    organization_user = models.ForeignKey(
+        "OrganizationUser",
+        on_delete=models.CASCADE,
+        related_name="flow_assistant_conversations",
+    )
+    title = models.CharField(max_length=200, blank=True, default="")
+    messages = models.JSONField(default=list)
+    started_at = models.DateTimeField(auto_now_add=True)
+    last_message_at = models.DateTimeField(auto_now=True)
```
VitaliiPovoroznykHYS: This filed will be updated every single time we update FlowAssistantConversation model including title and some other metadata stuff in future. It's better to set this to auto_now_add=True and update that timestamp manually when we have new message in messages

src/django_app/tables/models/flow_assistant_models.py
```diff
@@ -0,0 +1,48 @@
...
+        on_delete=models.CASCADE,
+        related_name="conversations",
+    )
+    organization_user = models.ForeignKey(
+        "OrganizationUser",
+        on_delete=models.CASCADE,
+        related_name="flow_assistant_conversations",
+    )
+    title = models.CharField(max_length=200, blank=True, default="")
+    messages = models.JSONField(default=list)
```
VitaliiPovoroznykHYS: Messages would be better as separate table, in this way we can add proper paginations for long chat histories to avoid big payloads, we can also add statistics and more robust filtering if needed and maybe some custom message types in future.

src/django_app/tables/services/flow_assistant/service.py
```diff
@@ -0,0 +1,1199 @@
...
+
+class LLMConfigInvalidError(Exception):
+    """Raised when the llm_config is misconfigured (e.g. unsupported provider)."""
+
+
+class ToolExecutionError(Exception):
+    """Raised when a tool function raises an unexpected exception."""
+
+
+# ── Tool registry ─────────────────────────────────────────────────────────────
```
VitaliiPovoroznykHYS: Tools should be separated to different file like flow_assistant/tools.py

src/django_app/tables/services/flow_assistant/service.py
```diff
@@ -0,0 +1,1199 @@
...
+    "get_session_detail": lambda graph_id, session_id, **_: _tools.get_session_detail(
+        graph_id, int(session_id)
+    ),
+    "get_session_messages": lambda graph_id, session_id, limit=50, **_: (
+        _tools.get_session_messages(graph_id, int(session_id), limit=int(limit))
+    ),
+}
+
+
+# ── Helpers ───────────────────────────────────────────────────────────────────
```
VitaliiPovoroznykHYS: separate helper.py file for this

src/django_app/tables/services/flow_assistant/service.py
```diff
@@ -0,0 +1,1199 @@
...
+
+_TITLE_MAX_CHARS = 50
```
VitaliiPovoroznykHYS: This goes into flow_assistant/constants.py

src/django_app/tables/services/flow_assistant/service.py
```diff
@@ -0,0 +1,1199 @@
...
+
+        # Build "Nodes in this flow" list — up to 30 entries, sorted by (type, id).
+        node_tuples: list[tuple[str, int, str]] = []
+        for node_type, model_cls, has_db_node_name in _NODE_TABLES:
+            fields = ["id", "node_name"] if has_db_node_name else ["id"]
+            for node in model_cls.objects.filter(graph_id=graph.pk).only(*fields):
+                node_tuples.append((node_type, node.pk, getattr(node, "node_name", "")))
+        node_tuples.sort(key=lambda t: (t[0], t[1]))
+
+        _MAX_NODES_IN_PROMPT = 30
```
VitaliiPovoroznykHYS, line 118: This variable should not be here, move to constants.py as well

src/django_app/tables/services/flow_assistant/service.py
```diff
@@ -0,0 +1,1199 @@
...
+        """Generate the persona system prompt from live flow metadata.
+
+        Always builds from current graph data — no caching so node additions
+        are visible immediately per conversation.start.
+        """
+        graph = flow_assistant.graph
+
+        # Node counts via single query per related manager
+        node_counts: dict[str, int] = {}
+        related_managers = [
```
VitaliiPovoroznykHYS: Would be nice to have this in a separate file as well, just because every single time we add a new node, we need to modify this list. 

In future we should put it somewhere separately cause this should be universal for all services when we want to access related names

src/django_app/tables/services/flow_assistant/service.py
```diff
@@ -0,0 +1,1199 @@
...
+                    f"  ... ({remainder} more — call the get_flow_overview tool to see all)"
+                )
+            nodes_section = "Nodes in this flow:\n" + "\n".join(lines)
+
+        now = timezone.now()
+        today_iso = now.date().isoformat()
+        yesterday_iso = (now - timedelta(days=1)).date().isoformat()
+        tomorrow_iso = (now + timedelta(days=1)).date().isoformat()
+
+        return (
```
VitaliiPovoroznykHYS: Would be much better if we have this as separate file that has just the prompt and accepts all the necessary fields as single dataclass and then just returns ready-to-use prompt

src/django_app/tables/services/llm_clients/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+    "TokenEvent",
+    "ToolCallEvent",
+    "ToolResultEvent",
+    "ToolSpec",
+    "UnsupportedLLMProviderError",
+    "get_llm_client",
+]
+
+# Lowercase provider-name → client class mapping
+_PROVIDER_MAP: dict[str, type[BaseLLMClient]] = {
```
VitaliiPovoroznykHYS: Why do we have hardcoded providers, when we already should support everything from light llm, which comes from our llm_config already?

src/django_app/tables/views/flow_assistant_views.py
```diff
@@ -0,0 +1,518 @@
...
+)
+from tables.services.flow_assistant.service import _request_cancel
+from tables.services.llm_clients.base import StructuredEvent
+from tables.services.rbac.organization_resolution import resolve_organization_user
+from tables.services.rbac.permissions import IsSuperadmin
+from tables.services.rbac.sse_ticket_service import SseTicketService
+from tables.utils.mixins import SSEMixin
+
+
+def _get_graph_or_404(graph_id: int):
```
VitaliiPovoroznykHYS, line 44: We have bult-in get_object_or_404 in django for this:

from django.shortcuts import get_object_or_404

src/django_app/tables/views/flow_assistant_views.py
```diff
@@ -0,0 +1,518 @@
...
+            return
+
+        # Build node index once for enriching tool_call SSE events.
+        node_index = await sync_to_async(_build_node_index)(graph_id)
+
+        service = FlowAssistantService()
+
+        try:
+            async for event in service.stream_reply(conversation, last_user_message):
+                if event.type == "token":
```
VitaliiPovoroznykHYS: This if/else tree looks bad, having separate handlers for each action sounds like better approach. Then we have easier time adding new events

VitaliiPovoroznykHYS (commented):
Future changes after demo

src/django_app/tables/services/flow_assistant/service.py
```diff
@@ -0,0 +1,1199 @@
...
+class LLMConfigMissingError(Exception):
+    """Raised when a FlowAssistant has no llm_config set."""
+
+
+class LLMConfigInvalidError(Exception):
+    """Raised when the llm_config is misconfigured (e.g. unsupported provider)."""
+
+
+class ToolExecutionError(Exception):
+    """Raised when a tool function raises an unexpected exception."""
```
ArturDrohomeretskyiHYS: we have file with exceptions: src\django_app\tables\exceptions.py, inherited from CustomAPIExeption 
batter to place it there

src/django_app/tables/services/flow_assistant/service.py
```diff
@@ -0,0 +1,1199 @@
...
+#
+# Inlined here rather than read at runtime to avoid brittle file-path coupling.
+# Adapted for the Flow Assistant: "Build mode" toggle and processTables action
+# are excluded (Code-Agent-specific); navigation guidance for "after creating /
+# modifying flows" is dropped (this assistant is read-only), but the openFlow /
+# openNode / refreshCache action verbs are retained as available actions.
+#
+# Update this string when the EpicChat response skill evolves.
+
+_RICH_FORMAT_GUIDANCE = """\
```
ArturDrohomeretskyiHYS: move this to smth like flow_assistant/prompts/ or flow_assistant/constants.py

src/django_app/tables/services/flow_assistant/service.py
```diff
@@ -0,0 +1,1199 @@
...
+
+
+# ── Cancel-flag helpers ───────────────────────────────────────────────────────
+#
+# A short-lived Redis key is set when the user hits the stop button.
+# `stream_reply` checks this flag at iteration boundaries and inside the
+# per-chunk loop to interrupt generation early.
+
+_CANCEL_KEY = "fa:cancel:{conv_id}"
+_CANCEL_TTL_SECONDS = 300
```
ArturDrohomeretskyiHYS: better to move on .env or src\django_app\tables\constants

src/django_app/tables/services/rbac/organization_resolution.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class OrganizationContextAmbiguous(CustomAPIExeption):
+    """Raised when the user belongs to multiple orgs and no X-Organization-Id header is set."""
+
+    status_code = 400
+    default_detail = (
+        "Multiple organization memberships; please specify X-Organization-Id header."
+    )
+    default_code = "organization_context_ambiguous"
```
ArturDrohomeretskyiHYS: we have file with exceptions: src\django_app\tables\services\rbac\rbac_exceptions.py
batter to place it there

src/django_app/tables/views/flow_assistant_views.py
```diff
@@ -0,0 +1,518 @@
...
+
+        serializer = SendMessageSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        message = serializer.validated_data["message"]
+
+        # Append user message to conversation so the stream view can access it
+        messages = list(conversation.messages)
+        messages.append({"role": "user", "content": message})
+        conversation.messages = messages
+        conversation.save(update_fields=["messages", "last_message_at"])
```
ArturDrohomeretskyiHYS: better to perform this logic inside FlowAssistantService()

src/django_app/tables/services/flow_assistant/tools.py
```diff
@@ -0,0 +1,1090 @@
...
+        "code": pc.code,
+        "entrypoint": pc.entrypoint,
+        "libraries": pc.get_libraries_list(),
+    }
+
+
+# ── internal helpers ──────────────────────────────────────────────────────────
+
+
+def _build_node_index(graph_id: int) -> dict[int, dict]:
```
ArturDrohomeretskyiHYS: this called internal helper but imported and used in flow_assistant_views.py

src/django_app/tables/views/flow_assistant_views.py
```diff
@@ -0,0 +1,518 @@
...
+    async def get_live_updates(self, pubsub):
+        """
+        Iterates FlowAssistantService.stream_reply and yields SSE events.
+        The pubsub parameter is ignored (we use an in-process async iterator).
+        """
+        from tables.services.flow_assistant import (
+            _build_node_index,
+            resolve_node_display_name,
+            resolve_subgraph_display_name,
+        )
```
ArturDrohomeretskyiHYS: better to make imports in top of file (general for PR)
dont import methods with  _ (supposed to be private)
and would be better to create the class with all helper methods or at least leave it as functions but in separate file 
smth like flow_asssitant/helpers.py 

now it imports from tables.services.flow_assistant due to src/django_app/tables/services/flow_assistant/__init__.py but it is kinda confused

src/django_app/tables/views/flow_assistant_views.py
ArturDrohomeretskyiHYS: agree 
also would be better to keep views as thin as possible and move business logic to service layer
this violates Open/Close Principle

src/django_app/tables/views/flow_assistant_views.py
ArturDrohomeretskyiHYS, line 1: general for file 
would be better to keep views as thin as possible. keep in views:
- permissions
- serialization
- calling dedicated method from service(s)
- return Response


+ make imports in top of file (if they do not create circular imports problem)

ArturDrohomeretskyiHYS (commented):
I only request changes for merge into main

src/django_app/tables/models/flow_assistant_models.py
```diff
@@ -35,7 +35,13 @@ class FlowAssistantConversation(models.Model):
...
     )
     title = models.CharField(max_length=200, blank=True, default="")
-    messages = models.JSONField(default=list)
+    # Legacy: kept for one release as a fallback. Writes go to FlowAssistantMessage rows.
+    # Drop in follow-up phase.
+    _messages_legacy = models.JSONField(
+        default=list,
+        db_column="messages",
+        editable=False,
+    )
```
ArturDrohomeretskyiHYS, line 44: do we actually need such a fallback?

VitaliiPovoroznykHYS (approved):

src/django_app/tables/services/flow_assistant/helpers.py
```diff
@@ -108,12 +108,81 @@ def _messages_for_llm(messages: list[dict]) -> list[dict]:
...
+    """
+    rows = list(
+        FlowAssistantMessage.objects.filter(
+            conversation_id=conversation_id
+        ).order_by("message_index")
+    )
+    if not rows:
+        # Transition fallback: no rows yet — read from legacy column.
+        conv = FlowAssistantConversation.objects.get(pk=conversation_id)
+        return list(conv._messages_legacy or [])
```
ArturDrohomeretskyiHYS, line 148: leave it now as is but remove in future (for main)

src/django_app/tables/migrations/0184_flow_assistant_message.py
ArturDrohomeretskyiHYS, line 1: it is OK
but better to use command `python manage.py makemigrations` for generating migration file and only after that edit it (python scripts, etc)

ArturDrohomeretskyiHYS (approved):

ArturDrohomeretskyiHYS:
Fix later spam in logs

PR #549, opened 2026-05-21
feat(EST-2885): Add more settings for flow`s import 

KarolinaKudrytskaHYS:
implement import flow options popover and settings management


src/django_app/tables/import_export/services/import_service.py
```diff
@@ -14,7 +14,11 @@ def __init__(self, registry: EntityRegistry):
         self.registry = registry
 
     def import_data(
-        self, export_data: dict, main_entity: str, preserve_uuids: bool = False
+        self,
+        export_data: dict,
+        main_entity: str,
+        preserve_uuids: bool = False,
```
VitaliiPovoroznykHYS: Move all import-related settings to separate dataclass ImportSettings

src/django_app/tables/import_export/strategies/graph.py
```diff
@@ -58,10 +59,12 @@ def extract_dependencies_from_instance(
     def export_entity(self, instance: Graph) -> dict:
         data = self.serializer_class(instance).data
         data["nodes"] = self._export_nodes(instance)
+        data["labels"] = [label.full_path for label in instance.labels.all()]
```
VitaliiPovoroznykHYS: It's better to move labels to separate Strategy

VitaliiPovoroznykHYS (changes requested):

src/django_app/tables/import_export/services/import_service.py
```diff
@@ -9,13 +10,26 @@
 from tables.import_export.constants import DEPENDENCY_ORDER
 
 
+@dataclass
+class ImportSettings:
+    preserve_uuids: bool = False
+    replace_existing: bool = False
+    import_labels: bool = True
```
VitaliiPovoroznykHYS: by default should be False

src/django_app/tables/import_export/services/import_service.py
```diff
@@ -62,7 +76,13 @@ def _resolve_import_order(self, export_data: dict) -> List[str]:
         return sorted_keys
 
     def _import_single_entity(
-        self, entity_data, entity_type, strategy, id_mapper, is_main, **kwargs
```
VitaliiPovoroznykHYS, line 65: keep **kwargs and ImportSettings dataclass

src/django_app/tables/import_export/services/import_service.py
```diff
@@ -72,6 +92,7 @@ def _import_single_entity(
 
         was_created = existing is None
 
+        kwargs = vars(settings) if settings is not None else {}
         instance = strategy.import_entity(entity_data, id_mapper, is_main, **kwargs)
```
VitaliiPovoroznykHYS: instead of this **kwargs just keep kwargs and add settings: ImportSettings to the base import_entity method of BaseImportStrategy

src/django_app/tables/import_export/strategies/label.py
```diff
@@ -0,0 +1,22 @@
+from tables.models import Graph
+from tables.models.label_models import Label
+
+
+class LabelStrategy:
```
VitaliiPovoroznykHYS: This should inherit from EntityImportExportStrategy

VitaliiPovoroznykHYS (changes requested):

src/django_app/tables/import_export/strategies/base.py
```diff
@@ -1,9 +1,12 @@
-from typing import Any, Optional
+from typing import Any, Optional, TYPE_CHECKING
 from abc import ABC, abstractmethod
 
 from tables.import_export.id_mapper import IDMapper
 from tables.import_export.enums import EntityType
 
+if TYPE_CHECKING:
```
VitaliiPovoroznykHYS: I don't think this is necessary. We can just import this

src/django_app/tables/import_export/strategies/base.py
```diff
@@ -42,14 +50,16 @@ def import_entity(
...
             existing = self.get_instance(existing_id)
             return existing
+
+        settings_kwargs = vars(settings) if settings is not None else {}
         if is_main:
-            return self.create_entity(data, id_mapper, **kwargs)
+            return self.create_entity(data, id_mapper, **settings_kwargs)
 
         existing = self.find_existing(data, id_mapper)
-        if existing and not is_main:
```
VitaliiPovoroznykHYS, line 49: Are you sure we don't need not is_main here?

src/django_app/tables/import_export/strategies/graph.py
VitaliiPovoroznykHYS, line 34: We should extract labels as dependency here

VitaliiPovoroznykHYS (changes requested):
You can check the docs for import/export. There is a section on how to add a new entity and just follow those steps

src/django_app/tables/import_export/strategies/base.py
OlgaMageriaHYS: If is_main=True the method returns on line 56 and never reaches line 58-59 at all. Lines 58-60 will be reached only when is_main=False, so additional check is not needed.

src/django_app/tables/import_export/strategies/label.py
```diff
@@ -0,0 +1,51 @@
...
+        old_parent_id = data.get("parent")
+        parent_id = (
+            id_mapper.get_or_none(EntityType.LABEL, old_parent_id)
+            if old_parent_id
+            else None
+        )
+        label, _ = Label.objects.get_or_create(name=data["name"], parent_id=parent_id)
+        return label
+
+    def attach_labels_to_graph(
```
VitaliiPovoroznykHYS: I would recommend to move this logic to GraphStrategy. Labels should not have data about entities they belong to. It's better to give every entity that will have labels handle their attachment in their own strategies.

VitaliiPovoroznykHYS (changes requested):

VitaliiPovoroznykHYS (approved):

VitaliiPovoroznykHYS (approved):

romanmasliak-byte (approved):

PR #551, opened 2026-05-21
Feature EST-2170 Implement Versioning and Prevent Overwrites

PavloNesterukHYS:
# Feature [EST-2170](https://hys-enterprise.atlassian.net/browse/EST-2170)

### Back - End:
- Add new property `save_version` to Graph table that acts like counter and used to compare if db value == user value. If equal - increment counter, if not - 409 Error
- Apply `save_version` check for graph PUT/PATCH, restore version (graph_versioning), and graph bulk save. 
- Modify serializers for endpoints listed above by including `save_version` field (see swagger for more details)
- Fix API tests for graph, bulk save and graph versioning


AndreyKrasnoyarskiy (approved):

src/django_app/tables/models/graph_models.py
```diff
@@ -62,6 +64,25 @@ class Graph(TimestampMixin, models.Model):
     epicchat_enabled = models.BooleanField(
         default=False, help_text="If 'True' -> flow is connected to EpicChat widget."
     )
+    save_version = models.IntegerField(default=1)
```
VitaliiPovoroznykHYS: use BigIntergerField

VitaliiPovoroznykHYS (changes requested):

VitaliiPovoroznykHYS (approved):

PR #552, opened 2026-05-21
fix(EST-2893): cleanup sandbox execution dirs on finish and startup

NikitaZalohaHYS:


VitaliiPovoroznykHYS (changes requested):
We also should remove sandbox_executions volume, we don't need that as persistent data, it should be okay if it resets every time

VitaliiPovoroznykHYS (approved):

PR #584, opened 2026-05-29
feature(EST-2987)/restore-localhost-webhooks

NikitaZalohaHYS:
- Restore ability to create localhost webhook
- Unify webhook logic for nodes and twilio with separate webhook trigger layer


src/django_app/tables/serializers/model_serializers/node_serializers/trigger_serializers.py
```diff
@@ -44,49 +49,39 @@ class Meta(BaseGraphEntityMixin.Meta):
...
-        wt = data.get("webhook_trigger")
-        if isinstance(wt, int):
-            self._webhook_trigger_id = wt
-            data = data.copy()
-            data["webhook_trigger"] = None
-        else:
-            self._webhook_trigger_id = None
-        return super().to_internal_value(data)
+    def _wait_for_tunnel_url(self, webhook_trigger):
+        from tables.services.webhook_trigger_service import WebhookTriggerService
```
VitaliiPovoroznykHYS: Move imports to top of the file

src/django_app/tables/serializers/utils/mixins.py
```diff
@@ -248,10 +253,87 @@ def _sync_tools(self, instance: Model, fk_to_instance: str, tool_ids: list[str])
...
 
 class WebhookCreationMixin:
+    @transaction.atomic
     def _get_or_create_webhook_trigger(self, data):
         path = data.get("path")
-        ngrok_conf = data.get("ngrok_webhook_config")
+        provider_type = data.get("provider_type")
+
+        try:
+            trigger = WebhookTrigger.objects.get(path=path)
```
VitaliiPovoroznykHYS: In this case can .get() return more then 1 entry? If yes - we should either change to filter or handle MultipleObjectsReturned exception

src/django_app/tables/serializers/base_serializers.py
```diff
@@ -1,17 +1,109 @@
...
+    def to_representation(self, instance):
+        rep = super().to_representation(instance)
+        ngrok = getattr(instance, "ngrok", None)
+        localhost = getattr(instance, "localhost", None)
+        rep["ngrok_config"] = NgrokConfigInlineSerializer(ngrok).data if ngrok else None
+        rep["localhost_config"] = (
+            LocalhostConfigInlineSerializer(localhost).data if localhost else None
+        )
+        try:
+            from tables.services.webhook_trigger_service import WebhookTriggerService
```
VitaliiPovoroznykHYS, line 92: Move import to the top

src/django_app/tables/serializers/base_serializers.py
```diff
@@ -1,17 +1,109 @@
...
+
+            rep["live_url"] = WebhookTriggerService().get_tunnel_url_for_trigger(
+                instance
+            )
+        except Exception:
```
VitaliiPovoroznykHYS, line 97: can we at least log that exception when it happens?

VitaliiPovoroznykHYS (changes requested):
Migrations are not merged. There are two 0190 migrations.

VitaliiPovoroznykHYS (approved):

src/django_app/tables/serializers/utils/mixins.py
```diff
@@ -248,10 +254,99 @@ def _sync_tools(self, instance: Model, fk_to_instance: str, tool_ids: list[str])
...
+                    f"WebhookTrigger with path={path!r} disappeared during lookup."
+                )
+            created = False
+
+        if not created and trigger.provider_type != provider_type:
+            # Provider changed — delete the old config to avoid orphan
+            if trigger.provider_type == ProviderType.NGROK:
+                NgrokWebhookConfig.objects.filter(trigger=trigger).delete()
+            elif trigger.provider_type == ProviderType.LOCALHOST:
+                LocalhostWebhookConfig.objects.filter(trigger=trigger).delete()
```
VitaliiPovoroznykHYS, line 381: Add logging, so when we delete a config here, at least we can see that clearly in logs

src/django_app/tables/serializers/base_serializers.py
```diff
@@ -1,17 +1,113 @@
...
+
+        if new_provider == ProviderType.NGROK and ngrok_data:
+            NgrokWebhookConfig.objects.update_or_create(
+                trigger=instance, defaults=ngrok_data
+            )
+            LocalhostWebhookConfig.objects.filter(trigger=instance).delete()
+        elif new_provider == ProviderType.LOCALHOST and localhost_data:
+            LocalhostWebhookConfig.objects.update_or_create(
+                trigger=instance, defaults=localhost_data
+            )
```
VitaliiPovoroznykHYS, line 79: This code leaves orphan config when triggered, since the deletion of orphans was handled only in 
`WebhookCreationMixin -> def _get_or_create_webhook_trigger(self, data):`

So basically PATCH leaves orphan when we change trigger.

VitaliiPovoroznykHYS (changes requested):

VitaliiPovoroznykHYS (approved):

PR #593, opened 2026-06-04
Feature/est 2423 resource scoping to organizations

ArturDrohomeretskyiHYS:


src/django_app/tables/models/graph_models.py
```diff
@@ -84,6 +85,14 @@ def increment_version_if_current(cls, pk: int, expected: int) -> int:
             raise GraphSaveVersionConflictError(current_version=current)
         return expected + 1
 
+    class Meta(OrgScopedModel.Meta):
+        abstract = False
```
BogdanTimchenkoHYS, line 89: abstract is not inherited, so you don't need to define it in the child and set false.

src/django_app/tables/models/llm_models.py
```diff
@@ -105,7 +114,7 @@ def delete(self, *args, **kwargs):
         return result
 
 
-class RealtimeModel(models.Model):
+class RealtimeModel(OrgScopedModel, models.Model):
```
BogdanTimchenkoHYS, line 117: Can inherit only OrgScopedModel without models.Model.

src/django_app/tables/models/python_models.py
```diff
@@ -13,23 +14,88 @@ def get_libraries_list(self):
...
+        ]
+
+    def get_tool_config_fields(self) -> dict[str, "PythonCodeToolConfigField"]:
+        if hasattr(self, "prefetched_config_fields"):
+            return {field.name: field for field in self.prefetched_config_fields}
+
+        return {
+            field.name: field
+            for field in PythonCodeToolConfigField.objects.filter(tool=self)
+        }
```
BogdanTimchenkoHYS: It's duplicating prefetch-cache logic of django.
Just use django natively.
`return {f.name: f for f in self.tool_fields.all()}`
If `.prefetch_relate('tool_fields') is called then '.all()` return the cashed result. Otherwise django executes ORM query.

src/django_app/tables/models/python_models.py
```diff
@@ -13,23 +14,88 @@ def get_libraries_list(self):
...
+        default=FieldType.STRING,
+    )
+    required = models.BooleanField(default=True)
+    secret = models.BooleanField(default=False)
+
+    class Meta:
+        unique_together = (
+            "tool",
+            "name",
+        )
```
BogdanTimchenkoHYS: Preferably use the UniqueConstraint because unique_together will be deprecated in future django version.

src/django_app/tables/models/python_models.py
```diff
@@ -13,23 +14,88 @@ def get_libraries_list(self):
...
+    def get_field(self, name: str) -> PythonCodeToolConfigField:
+        if hasattr(self.tool, "prefetched_config_fields"):
+            for field in self.tool.prefetched_config_fields:
+                if field.name == name:
+                    return field
+            return None
+
+        return PythonCodeToolConfigField.objects.filter(
+            tool=self.tool, name=name
+        ).first()
```
BogdanTimchenkoHYS: The same as this [review - duplicating prefetch-cache logic](https://github.com/EpicStaff/EpicStaff/pull/593/changes#r3535046010)

BogdanTimchenkoHYS:
# Require request in the serializer context
There are many cases where the `.validate_<something>` or `__init__` methods of serializers require a request from the context. Simply create a mixin (e.g. `RequireRequestSerializerMixin`) for the serializers that provides access to the request. 

# Complex and unclear contract for `ViewSet` and `APIView`
Currently, every developer has to pass the following to views every time:
- `OrgScopedServiceViewSetMixin`
- `HasOrgPermission`
- `rbac_resource_type`
- `rbac_action_map` (and don’t forget to pass in the default values).
We need to define a clear contract by creating our own base classes, `OrgViewSet` and `OrgApiView`. This way, the developer only needs to define the `rbac_action_map`. If there are custom endpoints and `rbac_resource_type`, there’s no need to pass any default RBAC resource types every time.

PavloNesterukHYS (approved):
Everything looks fine by me. 

*Checked the graph versioning and partially import/export (things that's related to the versioning)*

AndreyKrasnoyarskiy (approved):

VitaliiPovoroznykHYS (approved):

PR #597, opened 2026-06-08
Feature/est 3003 communication

BogdanTimchenkoHYS:
# What

New `src/shared/communication/` package — a transport-agnostic `Producer`/`Consumer` that splits message metadata (sent via a *broker*) from large payloads (kept in a *storage*), both behind interfaces and injected via DI. Ships a Redis broker and Redis + MinIO storages.

# Why

Services share one Redis pub/sub channel for metadata and payload: large payloads are duplicated per subscriber (RAM/OOM), block Redis on big publishes, and the business code is wired directly to the Redis client (hard to test, impossible to swap transport).

# How

`Producer` offloads payloads above a size threshold to the storage and sends only `{id, is_used_storage}` through the broker; `Consumer` rehydrates them on receive. Business code depends only on `AbstractBroker`/`AbstractStorage`; concrete backends are wired at startup. Covered by unit tests (fakeredis/mocks) and integration tests (real Redis + MinIO via testcontainers).

# Side Effects

- **Minimal implementation** — only the metadata/payload split with a Redis broker and Redis/MinIO storages. Reliable delivery, idempotency, extra backends and service migration are out of scope for this PR.
- Delivery is still fire-and-forget (Redis pub/sub); payload cleanup is best-effort — an offloaded payload is removed once the consumer advances past the message, otherwise it relies on storage TTL/lifecycle.

# Relates

[EST-3003](https://hys-enterprise.atlassian.net/browse/EST-3003) — Microservice Communication Optimization

### DIFF
--- a/src/shared/communication/brokers/redis_.py
+++ b/src/shared/communication/brokers/redis_.py
@@ -1,0 +1,53 @@
+import json
+from typing import AsyncIterable, Any, Iterable
+
+from redis import Redis as SyncRedis, RedisError
+from redis.asyncio import Redis as AsyncRedis
+
+from ..errors import BrokerOperationError
+from ..error_handler import handle_error
+from .abstract import AbstractBroker
+
+
+class RedisPubSubBroker(AbstractBroker):
+    """Redis Pub/Sub implementation of Broker.
+
+    Args:
+        url: Redis connection URL (e.g. redis://host:6379/0).
+    """
+
+    def __init__(self, url: str):
+        self._sync_client = SyncRedis.from_url(url)
+        self._async_client = AsyncRedis.from_url(url)
+
+    def send(self, channel: str, data: dict[str, Any]):
+        with handle_error(RedisError, BrokerOperationError, "send", channel):
+            raw_data = json.dumps(data)
+            self._sync_client.publish(channel, raw_data)
+
+    async def asend(self, channel: str, data: dict[str, Any]):
+        with handle_error(RedisError, BrokerOperationError, "asend", channel):
+            raw_data = json.dumps(data)
+            await self._async_client.publish(channel, raw_data)
+
+    def receive(self, channel: str) -> Iterable[dict[str, Any]]:
+        with handle_error(RedisError, BrokerOperationError, "receive", channel):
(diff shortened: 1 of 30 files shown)


src/shared/communication/producer.py
```diff
@@ -0,0 +1,62 @@
+import json
+
+from .message import Message
+from .brokers import AbstractBroker
+from .storages import AbstractStorage
+
+
+PAYLOAD_SIZE_THRESHOLD = 1024**2  # 1MB
```
ArturDrohomeretskyiHYS: name it DEFAULT_PAYLOAD_SIZE_THRESHOLD (it can be overridden during initialization Producer) and move to .env,
change comment or change code: 2mb or 1mb

src/shared/communication/brokers/redis_.py
```diff
@@ -0,0 +1,53 @@
+import json
+from typing import AsyncIterable, Any, Iterable
+
+from redis import Redis as SyncRedis, RedisError
+from redis.asyncio import Redis as AsyncRedis
+
+from ..errors import BrokerOperationError
+from ..error_handler import handle_error
```
ArturDrohomeretskyiHYS: consider to use absolute imports

src/shared/communication/storages/minio_.py
```diff
@@ -0,0 +1,85 @@
+import asyncio
+import io
+
+from minio import Minio, S3Error
+
+from ..errors import StorageOperationError
+from ..error_handler import handle_error
+from .abstract import AbstractStorage
```
ArturDrohomeretskyiHYS: consider to use absolute imports

src/shared/communication/producer.py
```diff
@@ -0,0 +1,65 @@
+import json
+from os import getenv
+
+from communication.message import Message
+from communication.brokers import AbstractBroker
+from communication.storages import AbstractStorage
+
+
+DEFAULT_PAYLOAD_SIZE_THRESHOLD = int(
```
ArturDrohomeretskyiHYS, line 9: declare variable here: src\\.env.example

ArturDrohomeretskyiHYS (approved):

PR #607, opened 2026-06-11
AEPF-52: When registering, add the clown emoji to the end of your username. This should only work from April 1st, and should also work on both sides: BE and FE.

NikitaZalohaHYS:
Completed stories:

- BE: Append clown emoji to organization name during registration on April 1st UTC
- FE: Display clown emoji appended to all organization names on April 1st UTC


ArturDrohomeretskyiHYS:
Please don't

PR #656, opened 2026-06-24
1682 knowledge statuses transfering

VasylCherkesHYS:
In this branch i add statuses to the knowledge service.

Now every document has a status. It can be chunking, chunked, indexing, completed, or failed. The service sets the status at each step.

The whole collection (RAG) also gets a status at the end:
completed — all documents are done.
warning — some documents are done, but some failed.
failed — all documents failed.
If a document fails, i save why. I store an error code and an error message (for example: chunking failed, embedding failed, or no chunks).

The prechunker makes preview chunks for one document.
First it sets the document status to chunking. Then it reads the file and makes the chunks.
When it works:
status is completed if the chunk config is the same as the last indexed one;
status is chunked if the config is new.
It saves the preview chunks.
When it fails (bad file or bad chunking), the status is failed, and we save the error code and message.

I also added a clear error when a collection has no documents.

I added tests for all these cases with mock data.


src/knowledge_NEW/models.py
```diff
@@ -97,13 +100,55 @@ class Document(Entity):
...
 
     @computed_field
     def extension(self) -> str:
         return Path(self.name).suffix
 
+    def config_same_with_snapshot(self) -> bool:
+        return (
+            self.last_indexing_config is not None
+            and self.config == self.last_indexing_config
+        )
```
BogdanTimchenkoHYS: `config_same_with_snapshot` is badly choice of name. Replace it with `is_required_reindex` and refactor the code
if the configurations match -> False
if the configurations differ -> True

src/knowledge_NEW/models.py
```diff
@@ -97,13 +100,55 @@ class Document(Entity):
...
+
+    def mark_completed(self) -> None:
+        self.status = DocumentStatusEnum.COMPLETED
+        self.last_indexing_config = self.config
```
BogdanTimchenkoHYS: The `last_indexing_config` requires a copy of the configuration, but you specified the original object.

src/knowledge_NEW/models.py
```diff
@@ -97,13 +100,55 @@ class Document(Entity):
...
+            self.status = DocumentStatusEnum.CHUNKED
+            self.completed_at = None
+        self.clear_error()
+
+    def clear_error(self) -> None:
+        self.error_code = DocumentErrorCode.NONE
+        self.error_message = None
+        self.failed_at = None
+
+    def clear_indexed_snapshot(self) -> None:
```
BogdanTimchenkoHYS: Replace `clear_indexed_snapshot` with `clear_last_indexing_config`.

src/knowledge_NEW/orchestrators/indexing/strategies/naive_indexer.py
```diff
@@ -102,14 +167,11 @@ async def index(self, request: IndexRequest, uow: SQLAlchemyUnitOfWork):
...
             async with uow:
-                await uow.naive_rag_repo.update_document(
-                    rag_id=request.rag_id,
-                    document=document,
-                )
                 if document.status == DocumentStatusEnum.COMPLETED:
                     await uow.naive_rag_repo.save_indexed_chunks(
                         document_id=document.id, chunks=document.indexed_chunks
                     )
+                await uow.naive_rag_repo.update_document(request.rag_id, document)
```
BogdanTimchenkoHYS: Why?) What does it change?) Restore it

src/knowledge_NEW/status_rules.py
```diff
@@ -0,0 +1,39 @@
+__all__ = [
+    "ERROR_MESSAGE_MAX_LENGTH",
+    "format_error_message",
+]
+
+ERROR_MESSAGE_MAX_LENGTH = 2000
+
+
+def _provider_error_message(exc: BaseException) -> str | None:
```
BogdanTimchenkoHYS: Replace `_provider_error_message`, as this name does not reflect the meaning.

src/knowledge_NEW/models.py
VasylCherkesHYS: method deleted

src/knowledge_NEW/models.py
```diff
@@ -97,13 +100,52 @@ class Document(Entity):
...
 
     @computed_field
     def extension(self) -> str:
         return Path(self.name).suffix
 
+    def is_required_reindex(self) -> bool:
+        return (
+            self.last_indexing_config is None
+            or self.config != self.last_indexing_config
+        )
```
BogdanTimchenkoHYS, line 119: if `last_indexing_config is None ` then it must returns `False`, because document must require indexing instead of reindexing  if it has no `last_indexing_config`.

src/knowledge_NEW/status_rules.py
```diff
@@ -0,0 +1,40 @@
...
+    `{"error": {"message": ...}}`.
+    """
+    body = getattr(exc, "body", None)
+    if isinstance(body, dict):
+        msg = body.get("message")
+        if not msg and isinstance(body.get("error"), dict):
+            msg = body["error"].get("message")
+        if msg:
+            return str(msg)
+    return None
```
BogdanTimchenkoHYS: Rename it to extract_provider_message

BogdanTimchenkoHYS (approved):

PR #662, opened 2026-06-30
Chore/est 3203 ruff and mypy

BogdanTimchenkoHYS:
### What
- Adds pre-commit–driven static analysis for the Python backend: ruff (lint + format)
  via `.ruff.toml` and mypy type-checking via `.mypy.ini`.
- Rewrites the existing `.pre-commit-config.yaml`, which was effectively non-functional.

### Why
- The backend had no enforced formatting / linting / type-checking gate, so style drift
  and untyped code landed unchecked.
- The pre-existing pre-commit config didn't actually guard anything: its `exclude`
  matched nothing nested, hook revisions were years stale, and the install comments
  were broken.

### How
- mypy runs in a deliberate "soft start": it checks untyped function bodies and forbids
  *partially* annotated functions, but does **not** require full annotations
  (`disallow_untyped_defs` left off). `warn_return_any` is intentionally disabled —
  with `ignore_missing_imports`, stubless libs (Django, …) return `Any` and would flood
  every function that forwards their result.
- Exclusions live in the **global** top-level pre-commit `exclude` using `(^|/)…/`, so
  `migrations/` and `libraries/` are skipped at any depth and filtered *before* any hook
  sees them — necessary for mypy, whose own config `exclude` is bypassed when pre-commit
  passes explicit file paths.
- ruff and mypy use standalone config files rather than `pyproject.toml` sections; ruff
  additionally repeats the excludes in `extend-exclude` so direct `ruff` runs match.

### Side Effects
- First run surfaces existing lint/type violations across the backend and can block
  commits; `ruff --fix` and `ruff-format` auto-modify files (double quotes, 100 cols) —
  expect a reformat churn on the initial pass.
- Contributors must install the hooks locally (`pip install pre-commit && pre-commit
  install`) for the gate to take effect.
- `.mypy.ini` defines no `exclude`, so a manual `mypy .` from the root (outside
  pre-commit) descends into `migrations/`, `libraries/`, and `tests/` — only the
  pre-commit path is filtered.

### Relates
[EST-3203](https://hys-enterprise.atlassian.net/browse/EST-3203) - Improve ruff config and add mypy

## Checklist

- [+] I have signed the **Contributor License Agreement (CLA)**. The CLA bot will comment on this PR — comment `I have read the CLA Document and I hereby sign the CLA` to sign. **This PR cannot be merged until the CLA is signed.**
- [+] My code follows the project's style and conventions.
- [+] I have tested my changes.
- [+] I have updated documentation where needed.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -1,0 +1,44 @@
+[mypy]
+# Docs: https://mypy.readthedocs.io/en/stable/config_file.html
+
+python_version = 3.12
+
+# Forbid partially annotated functions (some annotations present → all required).
+disallow_incomplete_defs = true
+
+# Type-check the bodies of unannotated functions (skipped by default).
+check_untyped_defs = true
+
+# Forbid untyped decorators (an untyped decorator erases the wrapped function's type).
+disallow_untyped_decorators = true
+
+# Warn when returning Any where a concrete type is declared.
+# DISABLED: with ignore_missing_imports, stubless libs (django, …) return Any and flood this.
+# warn_return_any = true
+
+# Warn about `# type: ignore` that no longer suppresses anything.
+warn_unused_ignores = true
+
+# Warn about cast() to a type the expression already has.
+warn_redundant_casts = true
+
+# Warn about per-module config sections that matched no files (catches typos).
+warn_unused_configs = true
+
+# Warn about statically unreachable code.
+warn_unreachable = true
+
+# Don't error on imports without stubs — treat such modules as Any.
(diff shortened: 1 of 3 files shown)


.mypy.ini
ArturDrohomeretskyiHYS, line 1: let's merge mypy config in the next release.

VitaliiPovoroznykHYS (approved):

ArturDrohomeretskyiHYS (approved):

PR #681, opened 2026-07-05
EST-3285: Tool Library Parity: implement Claude Code tool set for EpicStaff agents

NikitaZalohaHYS:
Vinik Pooh agent working on EST-3285. Draft — work in progress.


src/django_app/tables/views/views.py
```diff
@@ -624,6 +666,337 @@ def post(self, request, *args, **kwargs):
...
+        summary="Cancel a pending human-in-the-loop decision (timeout cleanup)",
+        request=CancelSessionDecisionSerializer,
+    )
+    def post(self, request, session_id, *args, **kwargs):
+        serializer = CancelSessionDecisionSerializer(data=request.data)
+        if not serializer.is_valid():
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+        try:
+            session = Session.objects.get(id=session_id)
```
NikitaZalohaHYS: **[Critical] Lost-update race: cancel can silently discard a just-committed human answer.** `CancelSessionDecisionView.post` reads the session with a plain unlocked `Session.objects.get(id=session_id)` (line 936) and later does a full-row `session.save()` (line 960, no `update_fields`), while `AnswerSessionDecisionView.post` correctly wraps the same read-modify-write in `transaction.atomic()` + `select_for_update()` (lines 859-861) precisely to guard this race. Interleaving is realistic: `wait_for_decision_tool` fires `_cancel_decision()` on timeout (main.py lines 229-237) while a user may be submitting an answer. Sequence: Answer (holding the lock) commits the answer and flips status to `run`; Cancel's unlocked SELECT still sees the pre-answer snapshot (`status=wait_for_user`, no answer), passes its guard (lines 947-951), then saves — popping `"decision"` from `status_data` and overwriting the row, silently discarding the recorded answer. The docstring (lines 918-924) acknowledges this exact race but the code doesn't mitigate it. Fix: use the same `transaction.atomic()` + `select_for_update()` pattern as Answer and re-check `decision.get("answer") is not None` inside the locked block.

src/shared/epicstaff_storage/storage.py
```diff
@@ -248,6 +311,68 @@ def delete(self, path: str) -> None:
...
+        batch_size = 1000
+        for i in range(0, len(keys), batch_size):
+            batch = keys[i : i + batch_size]
+            try:
+                client.delete_objects(  # type: ignore[attr-defined]
+                    Bucket=bucket,
+                    Delete={"Objects": [{"Key": key} for key in batch]},
+                )
+            except Exception as error:
+                self._handle_client_error(error, path)
```
NikitaZalohaHYS, line 435: **[Critical] `delete_folder` partial failure leaves S3 and the storage index permanently inconsistent.** The mutation-recording loop (lines 373-374) runs only after ALL delete batches succeed. If `client.delete_objects` (line 366) raises on batch N, `_handle_client_error` (line 371) always re-raises, so keys deleted in batches 1..N-1 are gone from S3 but **no mutations are recorded for any of them** — the Django `StorageFile` sync never removes those rows, leaving ghost index entries for deleted files. Compounding this: boto3's `delete_objects` returns HTTP 200 with per-key failures in `response["Errors"]`, and the response is ignored entirely (lines 366-369), so silently failed keys also get no correct accounting. Fix: record mutations per successful batch (inside the loop, e.g. try/finally), and check `response.get("Errors")`. Note the test suite can't catch this: `FakeS3Client.delete_objects` (tests/fakes.py lines 80-88) never simulates errors — a failure-injection test (fail on batch 2 of 2000+ keys, assert batch-1 mutations recorded) would cover it.

src/shared/tools/shell_exec_tool/main.py
```diff
@@ -0,0 +1,246 @@
...
+        text=True,
+        cwd=str(_working_root()),
+    )
+    if hasattr(os, "setsid"):
+        popen_kwargs["start_new_session"] = True
+
+    process = subprocess.Popen(command, **popen_kwargs)
+
+    try:
+        output, _ = process.communicate(timeout=timeout_seconds)
```
NikitaZalohaHYS: **[Critical] Unbounded output can OOM / disk-fill the single shared sandbox container.** Foreground mode buffers the ENTIRE command output in memory via `process.communicate()` (line 91) before the 30,000-char cap is applied (line 109) — a runaway command (an LLM-generated `print` loop is enough, no malice needed) buffers gigabytes into the tool process. Background mode is worse: output is shell-redirected to the job file with no size limit (`> "{output_file}" 2>&1`, lines 159/174), so an unbounded job fills the shared savefiles volume. Since docker-compose runs a single `sandbox` container for all sessions (docker-compose.yaml line 264) with no memory limit configured, either path takes down tool execution for every session, not just the offender. Fix: stream stdout with an incremental cap (stop reading/kill after N bytes) and wrap background commands with an output size limiter (e.g. `head -c`), plus consider a memory limit on the sandbox service.

src/shared/tools/shell_job_result_tool/main.py
```diff
@@ -0,0 +1,75 @@
...
+        if not job_id or not job_id.strip():
+            return "Error: job_id argument is mandatory and was not given to the tool."
+
+        # job_id is generated by uuid.uuid4().hex (shell_exec_tool) — always
+        # exactly 32 lowercase hex characters. Reject anything else outright
+        # rather than trusting it as a path component.
+        if len(job_id) != 32 or not all(c in "0123456789abcdef" for c in job_id):
+            return f"Error: invalid job_id '{job_id}'."
+
+        job_dir = _working_root() / JOBS_DIR_NAME / job_id
```
NikitaZalohaHYS: **[Critical] Background job results have no session-ownership check — cross-session output disclosure.** `job_id` is validated for format only (32 lowercase hex, lines 37-38) and then resolved directly to `_working_root() / JOBS_DIR_NAME / job_id` (line 40). The `.jobs/` directory lives on the savefiles volume that docker-compose mounts into the single shared `sandbox` container for ALL sessions (docker-compose.yaml lines 264, 276), so any session that knows (or brute-forces — there's no rate limit or attempt tracking) another session's job id can read that session's command output, which routinely contains sensitive data. Related: line 49 `read_text()`s the whole uncapped output file into memory before truncation (line 53), so retrieving a huge job result can also OOM the shared container. Fix: namespace job dirs by session id (and require the caller's session id to match), and read at most N bytes of the output file.

src/django_app/tables/views/views.py
```diff
@@ -624,6 +666,337 @@ def post(self, request, *args, **kwargs):
...
+        session.status_data = status_data
+        session.status = Session.SessionStatus.RUN
+        session.save()
+
+        return Response(
+            {"cancelled": True, "status": session.status}, status=status.HTTP_200_OK
+        )
+
+
+class NotifyEmailView(APIView):
```
NikitaZalohaHYS, line 629: **[Major] `NotifyEmailView` has no throttling — usable as an authenticated mail relay.** Any authenticated caller can send unlimited emails to arbitrary external addresses with fully caller-controlled subject (200 chars) and body (1000 chars) via `send_mail`/`DEFAULT_FROM_EMAIL`. The docstring claims auth means "this can't be used as an open mail relay", but auth only excludes anonymous callers — a low-privilege member, leaked API key, or (since this endpoint is driven by `notification_tool`, i.e. LLM output) a prompt-injected agent can still spam at will from your domain, risking deliverability/blacklisting. The codebase already has the convention for exactly this: `LoginThrottle`/`PasswordResetRequestThrottle` in `tables/throttles.py` wired via `DEFAULT_THROTTLE_RATES`. Add a scoped throttle (per-user/per-org) before this ships.

src/django_app/tables/views/views.py
```diff
@@ -624,6 +666,337 @@ def post(self, request, *args, **kwargs):
...
+        decision_id = str(uuid.uuid4())
+        decision_payload = {
+            "decision_id": decision_id,
+            "question": question,
+            "options": options,
+            "allow_free_text": allow_free_text,
+            "answer": None,
+        }
+
+        session.status = Session.SessionStatus.WAIT_FOR_USER
```
NikitaZalohaHYS: **[Major] `OpenSessionDecisionView` never checks session status — can resurrect a finished/crashed session into `wait_for_user` forever.** Line 772 unconditionally sets `session.status = WAIT_FOR_USER` and overwrites `status_data["decision"]` (lines 773-776). Both sibling endpoints gate on status (`AnswerSessionDecisionView` line 869 returns 409 unless `WAIT_FOR_USER`; `CancelSessionDecisionView` no-ops otherwise), but Open has no guard against `status in {end, error, stop, expired}`. A stale or duplicate `wait_for_decision_tool` call against a session whose crew already finished/crashed silently masks the terminal state and leaves the session stuck in `wait_for_user` with a decision no crew is listening to answer. Gate on `session.status == RUN` (mirroring the other two endpoints) and return 409 otherwise.

src/shared/tools/s3_file_insert_tool/main.py
```diff
@@ -1,12 +1,29 @@
...
+
+    storage = EpicStaffStorage()
     try:
-        EpicStaffStorage().insert_lines(file_path, line_number, content)
-        return f"Content inserted at line {line_number} in {file_path} successfully."
-    except ValueError as e:
-        return str(e)
+        try:
+            existing = storage.read(file_path)
+            line_count = len(existing.split("\n"))
```
NikitaZalohaHYS: **[Major] Line count inflated for files with a trailing newline.** `len(existing.split("\n"))` counts the empty string after a trailing newline: `"a\nb\n"` reports 3 lines instead of 2, so an insert at line 4 is accepted when only 2 lines exist, and the out-of-range error message (line 17) reports the same wrong count to the caller. This is inconsistent with `s3_file_count_lines_tool`, which correctly uses `content.count("\n")`-based semantics. Use the same convention here (and add a trailing-newline test case — the current test uses `"a\nb"` and misses this).

src/shared/tools/s3_file_line_read_tool/main.py
```diff
@@ -1,13 +1,38 @@
...
     except PermissionError as e:
         return str(e)
-    except StorageSizeLimitError as e:
-        return str(e)
     except ValueError as e:
         return str(e)
+    except RuntimeError as e:
+        return str(e)
+
+    lines = content.split("\n")
```
NikitaZalohaHYS: **[Major] Same trailing-newline line-count bug as s3_file_insert_tool.** `len(content.split("\n"))` over-counts files ending in `\n` (`"a\nb\n"` → 3 "lines"), so `offset=3` validates and reads a phantom empty line on a 2-line file. Inconsistent with `s3_file_count_lines_tool`'s `count("\n")` semantics — a caller that counts lines first then reads by offset gets contradictory results between the two tools. Align on one counting convention across all three tools and add trailing-newline/empty-file test cases.

src/crew/services/graph/nodes/crew_node.py
```diff
@@ -61,20 +61,27 @@ async def execute(
...
                 "variables": state["variables"].model_dump(),
                 "state_history": state["state_history"],
             },
+            # Exposed as a bare global (globals().get("session_id")) to
+            # built-in python-code tools — e.g. subflow_tool uses it to link
+            # a sub-flow run to its caller via Session.parent_session and to
+            # walk the ancestor chain for a recursion guard. Additive only:
+            # existing tools ignore unused globals, so this is a no-op for
+            # every tool that doesn't read "session_id".
+            "session_id": self.session_id,
```
NikitaZalohaHYS, line 86: **[Major] Injected `session_id` global silently clobbers any user-defined `session_id` variable, and the behavior is untested.** `global_kwargs = {**input_, "state": {...}, "session_id": self.session_id}` means a flow input or tool-author static config literally named `session_id` (a plausible name in a session-oriented product) is silently replaced by the internal numeric id — and `RunPythonCodeService.run_code()` merges `{**global_kwargs, **additional_global_kwargs}` so this dict always wins there too. The wrong value is used silently, which is very hard to root-cause in production. Also: `subflow_tool`'s recursion guard depends on this injection, yet there is no `test_crew_node.py` covering `CrewNode.execute()` — a regression here degrades the recursion guard silently (subflow_tool treats a missing `session_id` as non-fatal, main.py lines 132-141). Recommend at minimum a collision-precedence decision made explicit (reserved-name validation or namespaced key) plus a unit test pinning the injection.

frontend/src/app/pages/running-graph/components/graph-messages/components/decision-message/decision-message.component.ts
```diff
@@ -0,0 +1,101 @@
...
+            .submitDecisionAnswer(this.sessionId, payload)
+            .pipe(takeUntilDestroyed(this.destroyRef))
+            .subscribe({
+                next: () => {
+                    this.submittedOptionIndex.set(optionIndex);
+                    this.submittedFreeText.set(freeText);
+                    this.isAnswered.set(true);
+                    this.isSubmitting.set(false);
+                },
+                error: (error) => {
```
NikitaZalohaHYS: **[Major] Error handler hides the routine 409 "already answered" case and invites futile retries.** Every failure collapses into "Failed to submit your decision. Please try again." — but the backend's `AnswerSessionDecisionView` returns 409 specifically when the decision was already answered, which happens routinely here: the component's local `isAnswered` signal resets on re-render (page reload / SSE reconnect), so an already-answered decision re-renders as interactive and re-submitting always 409s. In that state "Please try again" is wrong. Branch on `error.status`: on 409 render the answered/closed state; on 404 show "this decision is no longer active".

NikitaZalohaHYS (commented):
## Automated review — EST-3285 tool-library parity PR

**Verdict: request changes** — 4 Critical and 6 Major findings posted as inline comments; Minor findings aggregated below. (Submitted as COMMENT because GitHub does not allow REQUEST_CHANGES from the PR author's own account.)

### Minor findings
- `src/shared/models/variable_conversion.py:63-70` — per-property `input_type` is now validated against `VariableTypeInput` and raises `ValueError` on invalid values; previously-tolerated typos in existing `tool_data.yaml` files will now fail at load. Intentional hardening, but worth a changelog/release note.
- `src/shared/tools/web_fetch_tool/main.py:22` — `_ssrf_guard` calls `httpx.URL(url)` without try/except; `notification_tool`'s `_ssrf_guard` (lines 53-56) wraps the same call defensively. Align for consistency.
- s3 tool tests — missing edge-case coverage for trailing-newline/empty/unicode files in `test_s3_file_insert_tool.py` and `test_s3_file_line_read_tool.py` (would have caught the two line-count bugs above), and no `head`/`tail` trailing-newline case in `test_s3_bash_tool.py`.
- `src/shared/epicstaff_storage/tests/` — no failure-injection test for `delete_folder` (e.g. `delete_objects` failing on batch 2); `FakeS3Client` never errors, so the partial-failure Critical above is invisible to the suite.
- `decision-message.component.html:27` — options `*ngFor` lacks `trackBy`, unlike the sibling findings table which added `trackByIndex`; index-based trackBy is fine here.
- `findings-message.component.scss:8` — hardcodes `box-shadow: 0 4px 16px rgba(0,0,0,0.2)` while this same PR introduces `--shadow-message-card` with that exact value (and decision-message uses the token). Use the variable.
- Frontend tests — no spec for `DecisionAnswerService` or `DecisionMessageComponent` (submit/guard/error branches). Matches the folder's existing no-spec convention, but the decision flow has real branching logic; a small service-level spec is cheap insurance.
- `decision-message.component.ts` — component receives only `sessionId`, so if the session reaches a terminal state while a decision is pending, options/Submit stay enabled and submission fails generically. Consider passing session status down to disable the pending decision UI.
- `src/docker-compose.yaml` — consider a memory limit on the shared `sandbox` service; it currently has none, which is what upgrades the shell-tool output findings above to platform-wide impact.

### What checked out clean
Cross-layer contract checks (Contract Integrity rule) verified clean by two independent reviewers: wait_for_decision message/answer payloads and findings payloads match exactly across report_findings_tool → session_callback_factory → graph-session-message.model.ts, and the answer POST body/URL match urls.py/serializers. No [innerHTML]/XSS surface in the new templates. Orchestration tools show consistent recursion guards (depth 5), bounded fanout workers, guaranteed spawn cleanup in finally, strict job-id format validation, and timeout-cancel in wait_for_decision_tool.

src/shared/tools/glob_tool/main.py
```diff
@@ -0,0 +1,94 @@
...
+            if not match.is_file():
+                continue
+            mtime = match.stat().st_mtime
+        except OSError:
+            continue
+        files.append((match, mtime))
+    return files
+
+
+def main(pattern: str, path: str | None = None) -> str:
```
VitaliiPovoroznykHYS, line 41: Path traversal in this tool escapes the sandbox root; .. in the glob pattern isn't checked against permission boundaries like every sibling tool does

src/shared/tools/shell_exec_tool/main.py
```diff
@@ -0,0 +1,410 @@
+# Shell Exec Tool
```
VitaliiPovoroznykHYS: I think we should not add tool like this one. It's way too dangerous and give access to sandbox files basically, if understand correctly.

src/shared/tools/shell_job_result_tool/main.py
```diff
@@ -0,0 +1,188 @@
+# Shell Job Result Tool
```
VitaliiPovoroznykHYS: Remove this tool as well. Same reason as shell_exec_tool

src/shared/tools/subflow_tool/main.py
```diff
@@ -0,0 +1,266 @@
+# Sub-Flow Tool
```
VitaliiPovoroznykHYS: I don't really understand this tool. First of all recursion check? When creating sublow there is already validation for recursion anyway. Second, I was thinking we just want this tool to run certain flow, but not as subflow literally no?

VitaliiPovoroznykHYS (changes requested):

VitaliiPovoroznykHYS (approved):

PR #688, opened 2026-07-09
Hotfix/est 3240 knowledge crash

BogdanTimchenkoHYS:
### What
- The knowledge service survives Redis pub/sub drops instead of dying with the listener: subscriptions auto-resubscribe with backoff, and TCP keepalive surfaces dead connections.
- GraphRAG re-indexing no longer crashes litellm's logging worker: `build_index` runs on a single persistent event loop instead of a fresh one per job.

### Why
- A dropped pub/sub connection silently killed the indexing/chunking/search listeners — the service stopped processing work until a restart.
- Each indexing job spun up a new loop via `asyncio.run`, orphaning litellm's process-global logging worker on the closed loop; the second indexing flooded the logs with `bound to a different event loop` and churned the worker through a die-and-respawn cycle.

### How
- `RedisService.listen()` owns a resubscribe loop: on `ConnectionError`/`TimeoutError` it re-subscribes with 5→30s backoff, so all three consumers get recovery without duplicating the logic.
- Keepalive (idle 30s / intvl 10s / 3 probes) + 5s connect timeout — half-open connections surface as errors the resubscribe loop can act on, instead of hanging.
- `run_executor` retries transient worker failures (search/chunking 3×; indexing intentionally 1× — it's expensive and not idempotent).
- `main()` stops swallowing listener exceptions (`gather` without `return_exceptions`, re-raises): now that the resubscribe loop absorbs Redis drops, a propagated error is a real fault worth crashing on.
- `BackgroundEventLoop` hosts one asyncio loop on a daemon thread; indexing submits `build_index` via `run_coroutine_threadsafe`, keeping litellm's worker bound to a single loop for the process lifetime.

### Side Effects
- Messages published during a reconnect window are lost — pub/sub has no replay, so indexing/search requests arriving mid-drop are dropped rather than queued.
- After a drop, a listener may resume with up to ~30s delay (backoff cap) — requests in that window go unanswered.
- An unhandled non-Redis listener exception now takes the whole service down (previously swallowed) — recovery relies on a container restart.
- If `stop()` exceeds its 5s join timeout at shutdown with an in-flight indexing job, the loop is closed under a running task — that indexing can be interrupted on shutdown.

### Relates
- [EST-3240](https://hys-enterprise.atlassian.net/browse/EST-3240) - `Knowledge` service crash-loops (~every 2 min, exit 0) → RAG searches time out
- [EST-2986](https://hys-enterprise.atlassian.net/browse/EST-2986) - Graph RAG bug with running several event loops

## Checklist

### DIFF
--- a/src/knowledge/tests/services/test_background_event_loop.py
+++ b/src/knowledge/tests/services/test_background_event_loop.py
@@ -1,0 +1,59 @@
+import asyncio
+
+import pytest
+from litellm.litellm_core_utils.logging_worker import GLOBAL_LOGGING_WORKER
+
+from services.background_event_loop import background_loop
+
+
+@pytest.fixture
+def isolated_logging_worker():
+    saved = (GLOBAL_LOGGING_WORKER._queue, GLOBAL_LOGGING_WORKER._worker_task)
+    GLOBAL_LOGGING_WORKER._queue = None
+    GLOBAL_LOGGING_WORKER._worker_task = None
+    yield GLOBAL_LOGGING_WORKER
+    GLOBAL_LOGGING_WORKER._queue, GLOBAL_LOGGING_WORKER._worker_task = saved
+
+
+async def _emit_one_callback(worker):
+    async def callback():
+        return None
+
+    worker.ensure_initialized_and_enqueue(callback())
+    await asyncio.sleep(0.05)
+
+
+def _worker_crashed(worker) -> bool:
+    task = worker._worker_task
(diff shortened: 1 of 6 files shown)


src/knowledge/tests/services/test_background_event_loop.py
ArturDrohomeretskyiHYS, line 1: translate doc-strings into English

ArturDrohomeretskyiHYS (approved):

PR #690, opened 2026-07-13
feat(EST-1893-BE): implement SecretCharField for secure handling of API keys and tokens across serializers

VasylCherkesHYS:


src/django_app/tables/serializers/utils/secret_fields.py
```diff
@@ -0,0 +1,35 @@
...
+        kwargs.setdefault("trim_whitespace", False)
+        super().__init__(*args, **kwargs)
+
+    def to_representation(self, value):
+        if value in (None, ""):
+            return value
+        s = str(value)
+        if len(s) <= 8:
+            return f"********"
+        return f"{'*'*(len(s)-4)}{s[-self.visible_tail :]}"
```
ArturDrohomeretskyiHYS: value is None or value == "" - better

isinstance(value, str) guard before str(value)

calculate len(s) one time, it needed for both cases 

f-string is redundant to return plain str

ArturDrohomeretskyiHYS (approved):

ElenaPrygunHYS (approved):

PR #695, opened 2026-07-16
feat(EST-3488-BE): implement PythonCodeResult model and serializers, add organization and user tracking for executions

VasylCherkesHYS:
Problem
Execution results had no real access control: PythonCodeResult rows weren't tied to any organization or user, so anyone could read any execution result by its execution_id (as a stopgap the endpoint had been locked down to superadmin-only, which also blocked regular users from their own results). On top of that there was no execution status: the row was created only after the run finished, and returncode defaulted to 0, so an unfinished run looked exactly like a successful one.

What was done
1. Permissions

PythonCodeResult now belongs to an organization and a user (org, created_by, plus a link to the executed python_code), stamped at launch.
GET /api/python-code-result/{execution_id}/ is org-scoped: OrgScopedViewSetMixin + HasOrgPermission with resource type FLOWS. A user sees only results of their active organization, and only with the FLOWS read permission. A foreign execution_id returns 404. Superadmin is no longer required.
2. Execution statuses

The result row is created immediately on POST /api/run-python-code/ with status="pending".
When the sandbox finishes, the Redis listener sets completed (returncode = 0) or error (≠ 0) and finished_at.
New fields: status, created_at, finished_at; returncode is null until the run finishes.
What the frontend needs to change
The endpoint now returns 200 immediately (previously 404 until the run finished), so "first 200 = done" polling no longer works:

poll until status !== "pending"; judge success by status (completed/error), not by returncode;
returncode, result_data, finished_at are null while pending;
404 now means "no such execution_id in your organization";
the endpoint works for regular users with the FLOWS permission — no superadmin needed.

GET /api/python-code-result/{execution_id}/ 

Before: 
404 until the run finished; the first 200 meant "done", response had no status:

{ 
"execution_id": "...", 
"result_data": "...",
 "returncode": 0, 
"stderr": "", 
"stdout": ""
 }

Now: 
200 right away (the row exists from the moment of launch), poll the status field:

// while running
{ 
"execution_id": "...", 
"status": "pending", 
"result_data": null,
"stdout": "", 
"stderr": "",
 "returncode": null, 
"created_at": "2026-07-16T15:24:17Z", 
"finished_at": null }

// finished
{ 
"execution_id": "...", 
"status": "completed" | "error", 
"result_data": "...", "stdout": "...",
 "stderr": "...", "returncode": 0, 
"created_at": "...", 
"finished_at": "2026-07-16T15:24:19Z" 
}

Poll until status !== "pending"; success = "completed", failure = "error" (don't check returncode).
404 = no such execution_id in your active organization (wrong org or bad id) — not "still running".
Auth: regular user with FLOWS read permission in the active org (superadmin no longer required).


src/django_app/tables/views/views.py
```diff
@@ -697,7 +697,12 @@ def post(self, request):
                 }
             )
 
-        execution_id = run_python_code_service.run_code(python_code.id, variables)
+        execution_id = run_python_code_service.run_code(
+            python_code.id,
+            variables,
+            organization_id=org_id,
+            user=request.user,
+        )
```
ArturDrohomeretskyiHYS, line 667: use all named params here

ArturDrohomeretskyiHYS (approved):

VasylCherkesHYS:

KarolinaKudrytskaHYS:

ViraKalenykHYS (approved):

PR #732, opened 2026-08-11
feat/est 3322 audit trail integration

NikitaZalohaHYS:
## Checklist


romanmasliak-byte:
[x] I have read the CLA Document and I hereby sign the CLA
[x] I have signed the Contributor License Agreement (CLA). The CLA bot will comment on this PR — comment I have read the CLA Document and I hereby sign the CLA to sign. This PR cannot be merged until the CLA is signed.
[x] My code follows the project's style and conventions.
[x] I have tested my changes.
[x] I have updated documentation where needed.

src/auditor/Dockerfile.auditor
VitaliiPovoroznykHYS, line 1: Add non-root user

src/auditor/pyproject.toml
```diff
@@ -0,0 +1,48 @@
+[project]
+name = "auditor"
+version = "1.0.0"
+description = "EpicStaff audit trail service"
+authors = [
+    { name = "EpicStaff" },
```
VitaliiPovoroznykHYS, line 6: Add "Nikita Zaloha" as well

src/auditor/app/core/settings.py
VitaliiPovoroznykHYS, line 1: So I remember I mentioned pydantic_settings, but since that time we changed our approach to how we work with .env variables.

Let's remove pydantic_settings as dependency and just work with them the same way as we do in other apps, like in agent for example: `src\agent\settings.py`

src/auditor/app/db/redis_client.py
```diff
@@ -0,0 +1,22 @@
...
+
+    Call this once, during app startup (lifespan) - not per-request, and not
+    at import time. Ownership of the single instance for the app's lifetime
+    belongs to app.state via the repository factory, not this function.
+    """
+    logger.info(f"Connecting to Redis at {settings.REDIS_HOST}:{settings.REDIS_PORT}")
+    return Redis(
+        host=settings.REDIS_HOST,
+        port=settings.REDIS_PORT,
+        password=settings.REDIS_PASSWORD or None,
```
VitaliiPovoroznykHYS: remove `or None` we never allow redis connection without password

src/docker-compose.yaml
```diff
@@ -524,6 +609,8 @@ volumes:
   redis_data: {}
   code_instances: {}
   minio_data: {}
+  opensearch_data: {}
```
VitaliiPovoroznykHYS: Mark opensearch_data as external. Also after this - update readme for project setup to include opensearch_data volume creation. It's easier to manage volumes this way when it's more explicit.

src/shared/audit/session_audit_writer.py
VitaliiPovoroznykHYS, line 1: Put this file under `audit/writers/` define Base class and extend this one from the base. In this way we have clear way to add new domains like user_actions_writer. Most likely add_custom_message can be shared between this classes

VitaliiPovoroznykHYS (changes requested):
Currently the chosen architecture looks not very scalable. It's not clear how to add new audit options like user actions, agent actions, etc. I would suggest the next app structure (just as an example, you might choose something differently), but the core idea is that we have clear separation of domains and adding new entry is just creating new domain under `domains/` with all the same file structure. Basically, everything that is a module constant today becomes per-domain data. I would also put routes into domains, but that's up to you to decide. The point is that I want to see clear way to add new domains

```
src/auditor/app/
|- domains/
|  |- base.py              # AuditDomain descriptor + the protocols
|  |- registry.py          # DOMAINS: dict[str, AuditDomain]
|  |- sessions/
|  |  |- domain.py         # the AuditDomain instance
|  |  |- fields.py         # KNOWN_FIELDS, aliases, flat roots
|  |  |- expansion.py      # session->node->event tree scope
|  |  |- computed.py       # duration
|  |  \- mappings/0001_audit_events.json
|  |- user_actions/        # same five files, different content
|  \- agent_actions/
|- filtering/              # AST shape + parser - domain-free
|- repositories/
|  |- opensearch_repository.py   # (client, index, sort_keys, model)
|  \- compiler.py                # QueryCompiler(catalog) - a class
|- services/
|  |- search_pipeline.py   # the ONE search flow
|  \- export_job_service.py
|- controllers/
|  \- domain_router.py     # router factory, mounted once per domain
\- index_setup/runner.py   # loops registry, creates + applies additive changes
```
---
Also example for common descriptor

```py
@dataclass(frozen=True)
class AuditDomain(Generic[T]):
    name: str                            # "sessions" - registry key + URL segment
    event_model: type[T]                 # SessionAuditEvent
    index: IndexSpec                     # index name, mapping path, sort keys
    fields: FieldCatalog                 # whitelist + aliases + flat roots
    scoping: ScopingPolicy               # org / retention clause builder
    computed: tuple[ComputedField, ...]  # duration, ...
    expander: MatchExpander              # session tree, or NullExpander
    resource: str                        # RBAC ResourceType for token gating
```
---
There are also some hard-coded stuff, that might be shaped into small protocols like:

| Protocol | Replaces | Sessions | user_actions |
|---|---|---|---|
| `FieldCatalog` | `KNOWN_FIELDS`, `_STRUCTURED_FIELDS`, `DEEP_FILTER_ALIASES`, `_FLAT_OBJECT_ROOTS` | 20 fields, 6 aliases | own set |
| `IndexSpec` | `SESSION_AUDIT_EVENTS_INDEX`, `_MAPPING_PATH`, the sort in `_execute` | `event_time desc, id desc` | `occurred_at desc, id desc` |
| `ScopingPolicy` | `scoped_query()` | `org_id` + retention on `event_time` | same shape, own time field |
| `ComputedField` | all of `services/duration_filter.py` | `duration` | none |
| `MatchExpander` | `services/match_scope.py` | 3-level tree | `NullExpander` |

---

Collapse the duplicated flow into one object, used by both the search route and the export route:

```py
class SearchPipeline:
    def __init__(self, domain: AuditDomain, repo: AuditRepository): ...
    async def search(self, request, scope) -> Page: ...      # cursor-paged
    async def scan(self, request, scope) -> AsyncIterator:   # export streams
```

VitaliiPovoroznykHYS (commented):
Second round of review (backend only, at `24d3f4de`).

Thanks for the restructure. The domain layout, registry, per-domain router factory and the shared `SearchPipeline` for search and export are exactly the direction I asked for. A shared `event_time` / `org_id` contract across domains is fine: every audit event needs a timestamp and an org.

A few things still need work before merge:

**Affects code already on `main`**
- `0211_grant_org_admin_organizations_update` sets the built-in Org Admin `organizations` permission to `4`. `main` already grants `6` (0209, 0242), so on existing DBs this removes READ from every Org Admin. Please delete it.
- `THIRD-PARTY-NOTICES.md` no longer lists django_app runtime deps (Django, DRF, drf-spectacular, gunicorn, channels). Please re-sync all venvs and regenerate.
- `USERS` is back in `rbac_enums.py` (main renamed it to `memberships`), and the `resource_type` choices drift from both migration branches. The branch is far behind `main`. Please merge `main` and produce one clean `AlterField` and merge migration.
- `.github/scripts/deploy.sh` doesn't create the now-external `opensearch_data` volume, so the next deploy will fail.
- `SessionData.org_id` is required now, which breaks the existing crew and django test fixtures (`test_graph_builder.py`, `test_graph_session_manager_service.py`, `test_task_node_payload.py`).

**Auditor**
- `ScopingPolicy` adds the org term only when `org_id` is passed. It's the only tenancy gate in the service, so make `org_id` required and raise when it's missing. A missing `retention_days` claim should be rejected, not treated as unlimited.
- Export collects the full result set in memory and writes with blocking I/O inside an async task. `scan()` should stream pages, write incrementally, and enforce a row cap.
- Bulk ingest returns 200 even when OpenSearch rejects documents, so audit rows get lost silently. Surface the errors so the producer retries.
- Per-domain token gating (`resource`) is still a TODO. Either implement it now or fail at startup when more than one domain is registered.
- The domain-free parts aren't fully domain-free yet: the compiler hard-codes session free-text fields and `error.raw`, `matching.py` assumes every request has `match_scope`, and the expander and duration code use `DEFAULT_SCOPING` instead of `domain.scoping`.

**Other**
- `crew/main.py` imports the audit provider by a different module path, so shutdown drains an empty client instead of the real one.
- The audit queue in `shared/audit/client.py` is unbounded and retries 4xx responses.
- The Django HITL emission (`register_message`) has no callers. Remove it or wire it properly.
- The decision-table node now publishes to `graph:messages` and changes execution order. That's a separate change and should go in its own PR.
- The manager sweep should confirm the path is inside the export dir before `unlink()`/`glob()`.
- Tests: please add a `SearchPipeline` test, a two-domain test, and cross-org and cross-user negatives for the audit token view, filter presets and exports.

The first-round inline comments are addressed (non-root user, authors, settings, Redis password, external volume). The writer base class is only partly there: `add_custom_message` isn't actually shared yet.

PR #747, opened 2026-08-17
fix(EST-3143): stop persisting realtime connection_key to localStorage

ViraKalenykHYS:
## Description

Three independent, individually-small security findings from a review, bundled under EST-3143 — each is a "free upgrade" for some other exploit (XSS → session hijacking, open recon endpoint → infra mapping, unbounded fetch → worker DoS).

***1. Realtime `connection_key` no longer touches localStorage.*** It was written to `localStorage` right after `POST init-realtime/` and read back twice (once to verify, once by the WebSocket client) — readable by any XSS anywhere in the app. Now carried as a plain parameter through `ConsoleService → RealtimeClient → RealtimeAPI` instead; no storage of any kind. Also simplified the RxJS pipe in `console.service.ts` (removed an unnecessary outer mutable variable + `tap()`, now a single `switchMap()` off the POST response).

***2. `GET /api/tunnel-url/{unique_id}` (webhook service) now requires auth.*** It had zero authentication and disclosed live ngrok tunnel URLs for any guessable `unique_id`. Added a `verify_webhook_auth` FastAPI dependency checking an `X-Webhook-Auth` header against the (previously unused) `WEBHOOK_AUTH` setting, constant-time compare. Fails closed (401) whether the header is wrong/missing *or* `WEBHOOK_AUTH` isn't configured — both cases return the identical response so an unauthenticated caller can't fingerprint which one it is; the real reason is only logged server-side. Confirmed no internal caller exists (Django reads this data straight from Redis), so nothing else needed updating. Updated the stale `# TODO: REMOVE IT GODDAMMIT` comments on `WEBHOOK_AUTH` across `debug.env`/`.env.example`/`env.yaml` to explain it's now required.

***3. Added request timeouts to two tools that had none.*** `crewai_xml_search_tool` (`timeout=10`) and `crewai_scrape_element_from_website_tool` (`timeout=15`, same bug found while fixing the first) — previously a slow/hung remote server could block a worker indefinitely.

## Checklist


src/webhook/app/core/auth.py
NikitaZalohaHYS, line 1: All changes for `src/webhook/` are redundant. Endpoint `/api/tunnel-url/{unique_id}` will be removed under Webhook changes in feature/EST-1869-voice-settings. Also for Webhook security measures we have EST-1728

src/.env.example
NikitaZalohaHYS, line 1: Changes in envs also should be reverted

NikitaZalohaHYS (changes requested):

src/.env.example
ViraKalenykHYS: Reverted, thanks

src/webhook/app/core/auth.py
ViraKalenykHYS: Reverted the webhook auth changes

NikitaZalohaHYS (approved):

KarolinaKudrytskaHYS (approved):

PR #750, opened 2026-08-18
fix(EST-3705): enforce READ_ONLY at the DB connection level in natura…

NikitaZalohaHYS:
…l_language_to_sql_tool

The tool previously relied solely on a crud_policy sentence injected into the LLM's prompt to prevent writes -- a request the model could be talked out of, not an actual permission. Now the underlying SQLAlchemy engine registers a connect-event hook that puts the session itself into read-only mode (Postgres: SET SESSION CHARACTERISTICS AS TRANSACTION READ ONLY; MySQL: SET SESSION TRANSACTION READ ONLY) before any agent-generated SQL can run on it, so the database rejects writes regardless of what the model produces. Unsupported dialects fail loudly instead of silently falling back to prompt-only enforcement. The prompt text is kept only as a UX hint.

Adds explicit sqlalchemy dependency and a test file covering the per-dialect listener behavior and the unsupported-dialect failure path.

## Checklist

### DIFF
--- a/src/shared/tools/natural_language_to_sql_tool/main.py
+++ b/src/shared/tools/natural_language_to_sql_tool/main.py
@@ -1,8 +1,14 @@
 # Natural Language To SQL Tool
 
+from sqlalchemy import create_engine, event
 from langchain_community.utilities import SQLDatabase
 from langchain_community.agent_toolkits import create_sql_agent
 from langchain_openai import ChatOpenAI
+
+READ_ONLY_SESSION_STATEMENTS = {
+    "postgresql": "SET SESSION CHARACTERISTICS AS TRANSACTION READ ONLY",
+    "mysql": "SET SESSION TRANSACTION READ ONLY",
+}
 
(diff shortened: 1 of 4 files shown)


src/shared/tools/natural_language_to_sql_tool/main.py
```diff
@@ -1,27 +1,53 @@
...
+                cursor.execute(statement)
+                cursor.close()
+                dbapi_connection.commit()
+
+        return engine
+
     def _create_agent(self):
         # TODO chould we parametrize that? at least model?
         llm = ChatOpenAI(
             model="gpt-4o-mini", temperature=0, api_key=self.openai_api_key
```
ArturDrohomeretskyiHYS: not this feature, but model can not be hardcoded. Also it means this tool will not work with other provider except OpenAI

ArturDrohomeretskyiHYS (approved):

PR #751, opened 2026-08-18
feat(EST-3788): add soft delete option for entities

SofiiaDudnichenkoHYS:
## Checklist


VitaliiPovoroznykHYS (approved):

src/django_app/django_app/settings.py
```diff
@@ -33,6 +33,9 @@
 # SECURITY WARNING: keep the secret key used in production secret!
 DEBUG = os.getenv("DEBUG", "False").lower() in ("true", "1", "yes", "on")
 
+# Controls whether SoftDeleteMixin.delete() soft-deletes (mark inactive) or hard-deletes.
+SOFT_DELETE = os.getenv("SOFT_DELETE", "True").lower() in ("true", "1", "yes", "on")
```
VitaliiPovoroznykHYS: Set default to False

src/django_app/tables/models/base_models.py
```diff
@@ -129,30 +130,61 @@ class Meta:
...
+    Only the fields required for soft deletion. No delete() override —
+    a direct .delete() on a model that only has this mixin (no SoftDeleteMixin)
+    performs a normal, unconditional Django hard delete.
+    """
+
+    is_soft_deleted = models.BooleanField(
+        default=False, db_default=False, db_index=True
     )
-    is_active = models.BooleanField(default=True)
+    soft_deleted_at = models.DateTimeField(null=True, blank=True, db_index=True)
```
VitaliiPovoroznykHYS: A standard B-tree index on a boolean is virtually useless to the query planner, and creating these takes a heavy write lock on massive tables. Please drop `db_index` on both

src/django_app/tables/import_export/strategies/graph.py
```diff
@@ -156,7 +156,12 @@ def create_entity(self, data: dict, id_mapper: IDMapper, **kwargs) -> Graph:
         imported_uuid = import_data.pop("uuid", None)
         if preserve_uuids and imported_uuid:
             if replace_existing:
-                Graph.objects.filter(uuid=imported_uuid).delete()
+                # Queryset .delete() bypasses Model.delete() entirely (Django never
+                # calls instance delete() for queryset-level deletes), so it always
+                # hard-deletes regardless of settings.SOFT_DELETE. Delete per-instance
+                # instead so SoftDeleteMixin.delete() (and thus DeleteService) fires.
+                for existing_graph in Graph.objects.filter(uuid=imported_uuid):
+                    existing_graph.delete()
```
VitaliiPovoroznykHYS, line 159: Importing with `replace_existing` + `preserve_uuids` now returns a 500 error. Soft-deleted graphs keep their rows and UUIDs. Validation passes because DRF uses `ActiveManager`, but the `INSERT` hits the DB index and causes an unhandled `IntegrityError`.

**Fix:** Clear the UUID upon soft deletion.

src/django_app/tables/services/soft_delete.py
```diff
@@ -0,0 +1,452 @@
...
+        SoftDeleteFields always wins.
+
+        We do not care whether its FK says CASCADE,
+        SET_NULL, PROTECT, etc. If the dependent object
+        itself supports soft deletion, it gets soft-deleted.
+        """
+
+        if isinstance(child, SoftDeleteFields):
+            self.delete(child)
+            return
```
VitaliiPovoroznykHYS: This `SoftDeleteFields` check runs before `on_delete` is evaluated, overriding `SET_NULL` and breaking other flows. For example, soft-deleting Graph B will soft-delete `SubGraphNode` rows owned by Graph A. Graph A stays active but loses a node, and eventually crashes with a `DoesNotExist` error in `base_models.py:337` during validation/serialization.


**Fix:** Handle `SET_NULL`/`SET_DEFAULT`/`SET()` before the soft-delete branch, and switch `find_globally` to use `_base_manager` with an explicit soft-delete check.

src/django_app/tables/services/soft_delete.py
```diff
@@ -0,0 +1,452 @@
...
+                obj,
+                relation,
+            )
+
+            for child in children:
+                self._process_related_object(
+                    parent=obj,
+                    child=child,
+                    relation=relation,
+                )
```
VitaliiPovoroznykHYS, line 188: Cascade logic is running row-by-row in Python. It fetches `manager.all()` and issues a `save()` / `delete()` per row, recursing deeply. Deleting one Graph with thousands of sessions causes thousands of DB round trips in a single transaction and risks maxing out recursion depth. 

**Fix:** Batch this per model per level using `queryset.update(is_soft_deleted=True, soft_deleted_at=now)`.

src/django_app/tables/models/graph_models.py
VitaliiPovoroznykHYS, line 49: The `get_transitive_subflows` recursive CTE ignores soft deletes. Traversal passes right through soft-deleted nodes because there is no `is_soft_deleted` predicate.

src/django_app/tables/services/soft_delete.py
```diff
@@ -0,0 +1,452 @@
...
+
+    @staticmethod
+    def _get_reverse_relations(obj):
+        for relation in obj._meta.get_fields():
```
VitaliiPovoroznykHYS: `obj._meta.get_fields()` excludes hidden reverse relations (`related_name="+"`). As a result, relations like `schedule_trigger_node` in `session_models.py:127` are invisible to the walker.

**Fix:** Use `get_fields(include_hidden=True)` and query by `relation.field.name` when `get_accessor_name()` is None.

src/django_app/tables/services/soft_delete.py
```diff
@@ -0,0 +1,452 @@
...
+
+    @classmethod
+    @transaction.atomic
+    def delete(
+        cls,
+        obj: models.Model,
+        using: str | None = None,
+    ):
+        context = _DeleteContext(using=using)
+        context.delete(obj)
```
VitaliiPovoroznykHYS: The bare `@transaction.atomic` decorator opens the block on the default DB, while every write actually goes to `self.using`. 

**Fix:** Use `with transaction.atomic(using=using):` inside the method body.

---

```
@classmethod
def delete(cls, obj, using: str | None = None):
    with transaction.atomic(using=using):
        context = _DeleteContext(using=using)
        context.delete(obj)
```

src/django_app/tables/services/soft_delete.py
```diff
@@ -0,0 +1,452 @@
...
+                "guarantee referential integrity."
+            )
+
+        # ------------------------------------------------------
+        # CASCADE
+        # ------------------------------------------------------
+
+        if on_delete is models.CASCADE:
+            self.delete(child)
+            return
```
VitaliiPovoroznykHYS: The `on_delete` chain is a massive if/elif type-check that is hard to maintain and risks breaking if Django ever renames `django.db.models.SET`

**Fix:** Use a `{sentinel: handler}` dictionary map.

src/django_app/tables/models/base_models.py
```diff
@@ -129,30 +130,61 @@ class Meta:
...
+
+    objects = ActiveManager()
+    all_objects = models.Manager()
+
+    class Meta:
+        abstract = True
```
VitaliiPovoroznykHYS, line 180: Two fields (`is_soft_deleted` and `soft_deleted_at`) encode one state, but nothing keeps them consistent. 

**Fix:** add a database `CheckConstraint` to ensure they match.

src/django_app/tables/models/crew_models.py
VitaliiPovoroznykHYS, line 1: The mixin position varies across models. Django resolves `_default_manager` by MRO depth, meaning if another abstract base declares `objects`, filtering will silently disappear for models that list `SoftDeleteFields` later.

Same for `graph_models.py:959`

**Fix:** Standardize the mixin position and explicitly set `Meta.default_manager_name` and `base_manager_name`.

VitaliiPovoroznykHYS (changes requested):
**Please write documentation for this feature.**

---

**Scheduled triggers keep firing for deleted flows.** 

Because this uses raw SQL, the `ActiveManager` never applies, meaning cron jobs will start sessions on graphs the user already deleted. 

**Fix:** Update the query to `WHERE is_active = true AND is_soft_deleted = false`
**File: `schedule_trigger_repository.py` | Line 51**

---

VitaliiPovoroznykHYS (approved):

PR #755, opened 2026-08-19
Fix/EST-3794 fix python code tool removing

SofiiaDudnichenkoHYS:
## Checklist


src/django_app/tables/services/python_code_cleanup.py
```diff
@@ -0,0 +1,53 @@
...
+
+    exclude_model skips checking the model whose instance just triggered this
+    call via post_delete — its row is already gone, so it's excluded from the
+    ownership check without adding a subquery for it.
+    """
+    ids = {code_id for code_id in python_code_ids if code_id is not None}
+    if not ids:
+        return
+
+    relations = [r for r in OWNER_RELATIONS if r.related_model is not exclude_model]
```
ArturDrohomeretskyiHYS: This excludes all relations of the sender model, not just the row that was deleted. If two owners of the same model share a PythonCode, deleting one makes the guard see zero owners, so the code gets deleted — and since the owner FK is CASCADE, the second node gets deleted with it. Two PythonNodes on the same code, or CDT node A with pre_python_code=X and node B with post_python_code=X (that one excludes both columns at once).

Fix — drop the parameter and always check every relation:


```
    candidates = PythonCode.objects.filter(id__in=ids)
    candidates.exclude(_still_owned(OWNER_RELATIONS)).delete()
```
(and remove` *, exclude_model=None` from the signature on line 40 plus the paragraph about it in the docstring, and `exclude_model=sender` from the signal handler)

src/django_app/tables/services/python_code_cleanup.py
```diff
@@ -0,0 +1,53 @@
...
+                        **{relation.field.attname: OuterRef("id")}
+                    )
+                )
+            )
+            for relation in relations
+        ),
+    )
+
+
+def delete_python_code(python_code_ids: set[int], *, exclude_model=None) -> None:
```
ArturDrohomeretskyiHYS: set[int] -> Iterable[int | None]

src/django_app/tables/services/python_code_cleanup.py
```diff
@@ -0,0 +1,53 @@
...
+    ]
+
+
+OWNER_RELATIONS = tuple(_owner_relations())  # computed once, at import time
+
+
+def _still_owned(relations) -> Q:
+    """Q matching rows referenced by at least one of the given relations."""
+    if not relations:
+        return Q(pk__in=[])
```
ArturDrohomeretskyiHYS: Problem: this is consumed by exclude(), so "no relations discovered" ends up meaning "delete everything" — it fails open in the one case where we know least.

Fix — return an empty Q and let the caller bail. bool(Q()) is False, so the guard at the callsite is a one-liner:


```
    if not relations:
        return Q()
```
then in `delete_python_code`:


```
    owners = _still_owned(OWNER_RELATIONS)
    if not owners:
        return
```

src/django_app/tables/services/python_code_cleanup.py
```diff
@@ -0,0 +1,53 @@
...
+            Q(
+                Exists(
+                    relation.related_model.objects.filter(
+                        **{relation.field.attname: OuterRef("id")}
+                    )
+                )
+            )
+            for relation in relations
+        ),
+    )
```
ArturDrohomeretskyiHYS: Problem: only place in the app importing functools.reduce/operator.or_, and the generator-inside-reduce takes a beat to read.

Fix:


```
    condition = Q()
    for relation in relations:
        condition |= Q(
            Exists(
                relation.related_model.objects.filter(
                    **{relation.field.attname: OuterRef("id")}
                )
            )
        )
    return condition
```
(lets you drop lines 1-2, the reduce/or_ imports)

src/django_app/tables/services/python_code_cleanup.py
```diff
@@ -0,0 +1,53 @@
...
+
+def _owner_relations():
+    return [
+        relation
+        for relation in PythonCode._meta.related_objects
+        if relation.on_delete is models.CASCADE
+    ]
+
+
+OWNER_RELATIONS = tuple(_owner_relations())  # computed once, at import time
```
ArturDrohomeretskyiHYS: Problem: introspects the model registry at import time. Works today only because the module is reached through AppConfig.ready() — flow_control_serializers now imports it too, and the first import from models.py or apps.py gives AppRegistryNotReady. The comment also restates the code instead of the reason.

Fix — make it lazy, same one-time cost:


```
@cache
def owner_relations() -> tuple:
    """CASCADE reverse relations into PythonCode. Cached, not import-time, so
    importing this module never depends on the app registry being ready."""
    return tuple(
        relation
        for relation in PythonCode._meta.related_objects
        if relation.on_delete is models.CASCADE
    )
```
with `from functools import cache` up top, and callers switching from `OWNER_RELATIONS` to `owner_relations()`.

src/django_app/tables/services/python_code_cleanup.py
```diff
@@ -0,0 +1,53 @@
...
+    call via post_delete — its row is already gone, so it's excluded from the
+    ownership check without adding a subquery for it.
+    """
+    ids = {code_id for code_id in python_code_ids if code_id is not None}
+    if not ids:
+        return
+
+    relations = [r for r in OWNER_RELATIONS if r.related_model is not exclude_model]
+    candidates = PythonCode.objects.filter(id__in=ids)
+    candidates.exclude(_still_owned(relations)).delete()
```
ArturDrohomeretskyiHYS: Problem: no transaction, no locking. A request attaching a new owner between the EXISTS check and the cascade loses its node.

Fix — lock the candidate rows in a separate SELECT, since .delete() discards select_for_update():


```
    with transaction.atomic():
        locked_ids = set(
            PythonCode.objects.select_for_update()
            .filter(id__in=ids)
            .values_list("id", flat=True)
        )
        deleted, _ = (
            PythonCode.objects.filter(id__in=locked_ids).exclude(owners).delete()
        )
    return deleted
```

src/django_app/tables/services/python_code_cleanup.py
ArturDrohomeretskyiHYS, line 1: Problem: everything else in services/ is a class (SearchConfigService, PythonCodeToolCopyService, RagAssignmentService), and this is a module of functions with a module-level global.

Fix — tables/services/python_code_cleanup_service.py, folding in every suggestion above:


```
from functools import cache
from typing import Iterable

from django.db import models, transaction
from django.db.models import Exists, OuterRef, Q
from loguru import logger

from tables.models.python_models import PythonCode


class PythonCodeCleanupService:
    """Deletes PythonCode rows once no CASCADE owner references them.

    The owner FKs run owner -> PythonCode with on_delete=CASCADE, so the DB
    only cascades in the PythonCode-deleted direction; nothing reclaims a row
    when its last owner goes away. This service is that missing direction.
    """

    @classmethod
    @cache
    def owner_relations(cls) -> tuple:
        """CASCADE reverse relations into PythonCode, discovered once, lazily."""
        return tuple(
            relation
            for relation in PythonCode._meta.related_objects
            if relation.on_delete is models.CASCADE
        )

    @classmethod
    def owner_fields_by_model(cls) -> dict[type, list[str]]:
        owners: dict[type, list[str]] = {}
        for relation in cls.owner_relations():
            owners.setdefault(relation.related_model, []).append(
                relation.field.attname
            )
        return owners

    @classmethod
    def owned_by_any_owner(cls) -> Q:
        """Q matching PythonCode rows still referenced by at least one owner."""
        condition = Q()
        for relation in cls.owner_relations():
            condition |= Q(
                Exists(
                    relation.related_model.objects.filter(
                        **{relation.field.attname: OuterRef("id")}
                    )
                )
            )
        return condition

    @classmethod
    def delete_orphaned(cls, python_code_ids: Iterable[int | None]) -> int:
        """Delete the given PythonCode rows that have no owner left."""
        ids = {code_id for code_id in python_code_ids if code_id is not None}
        if not ids:
            return 0

        owners = cls.owned_by_any_owner()
        if not owners:
            return 0

        with transaction.atomic():
            locked_ids = set(
                PythonCode.objects.select_for_update()
                .filter(id__in=ids)
                .values_list("id", flat=True)
            )
            deleted, _ = (
                PythonCode.objects.filter(id__in=locked_ids).exclude(owners).delete()
            )

        if deleted:
            logger.debug(
                "Deleted {deleted} orphaned PythonCode rows", deleted=deleted
            )
        return deleted
```
`owned_by_any_owner()` being public also gives the cleanup command below something to reuse.

src/django_app/tables/signals/python_code_signals.py
```diff
@@ -0,0 +1,29 @@
...
+
+_OWNER_FIELD_NAMES_BY_MODEL = _owner_field_names_by_model()
+
+
+def _delete_owned_python_code(sender, instance, **kwargs):
+    field_names = _OWNER_FIELD_NAMES_BY_MODEL.get(sender)
+    if not field_names:
+        return
+    python_code_ids = {getattr(instance, field_name) for field_name in field_names}
+    delete_python_code(python_code_ids, exclude_model=sender)
```
ArturDrohomeretskyiHYS: Problem: fires per deleted instance, and each call isn't one query — exclude(...).delete() fetches the rows, then Collector walks all six CASCADE relations plus the SET_NULL on PythonCodeResult plus the secrets M2M. Ballpark ten queries per deleted node, so a graph with 200 python nodes goes from a handful of statements to a couple thousand.

Fix — collect during the delete, flush once at commit. Bulk save is already atomic (@transaction.atomic on _execute_writes), so this collapses a whole flow-save to one cleanup:


```
import threading

from django.db import transaction
from django.db.models.signals import post_delete

from tables.services.python_code_cleanup_service import PythonCodeCleanupService

_state = threading.local()


def _buffer(python_code_ids) -> None:
    pending = getattr(_state, "pending", None)
    if pending is None:
        pending = _state.pending = set()
        transaction.on_commit(_flush)
    pending.update(python_code_ids)


def _flush() -> None:
    pending, _state.pending = getattr(_state, "pending", set()), None
    PythonCodeCleanupService.delete_orphaned(pending)


def _collect_owned_python_code(sender, instance, **kwargs):
    field_names = PythonCodeCleanupService.owner_fields_by_model().get(sender)
    if not field_names:
        return
    _buffer({getattr(instance, name) for name in field_names})
```

src/django_app/tables/serializers/model_serializers/node_serializers/flow_control_serializers.py
```diff
@@ -341,6 +348,14 @@ def update(self, instance, validated_data):
             setattr(instance, attr, value)
         instance.save()
 
+        delete_python_code(detached_python_code_ids)
+        if detached_python_code_ids:
+            logger.debug(
+                "Detached and cleaned up PythonCode {ids} from CDT node {node_id}",
+                ids=sorted(detached_python_code_ids),
+                node_id=instance.id,
+            )
```
ArturDrohomeretskyiHYS: Problem: the log claims "Detached and cleaned up PythonCode {ids}" but the guard may have kept every one of them, and the if detached_python_code_ids: check duplicates the service's own if not ids: return.

Fix — let the service log the real count (it does, in the class above) and keep the callsite to one line:


```
        PythonCodeCleanupService.delete_orphaned(
            python_code_ids=detached_python_code_ids
        )
```

ArturDrohomeretskyiHYS (changes requested):
Duplicate cleanup left behind — graph_versioning/manager.py:687-715. _wipe_graph_children already does this by hand, and it's now both redundant (the .all().delete() loop fires the new signals first) and weaker: it collects only PythonNode/ConditionalEdge/WebhookTriggerNode, missing both CDT columns, and treats only PythonCodeTool as a co-owner — so a code shared with a node in another graph gets deleted and cascade-deletes that node.

Fix: delete the collection and the orphan block, keep just the wipe:


```
    def _wipe_graph_children(self, graph: Graph) -> None:
        """Wipe all graph related nodes. Orphaned PythonCode rows are reclaimed
        by the post_delete cleanup in tables.signals.python_code_signals."""
        for relation_name in _GRAPH_RELATION_NAMES:
            getattr(graph, relation_name).all().delete()
```
That drops the now-unused PythonCode / PythonCodeTool / PythonNode / ConditionalEdge / WebhookTriggerNode imports at the top of the module if nothing else uses them, and it needs a check in the versioning tests that restore still leaves no orphans.

SofiiaDudnichenkoHYS:

src/django_app/tables/services/python_code_cleanup_service.py
```diff
@@ -0,0 +1,71 @@
...
+
+        with transaction.atomic():
+            locked_ids = set(
+                PythonCode.objects.select_for_update()
+                .filter(id__in=ids)
+                .values_list("id", flat=True)
+            )
+            deleted, _ = (
+                PythonCode.objects.filter(id__in=locked_ids).exclude(owners).delete()
+            )
```
VitaliiPovoroznykHYS, line 68: `select_for_update().filter(id__in=ids)` locks rows in whatever order Postgres returns them. Two concurrent flushes with overlapping id sets can each grab a row the other needs and deadlock.

`_flush` runs on `on_commit` for every request that deletes a python-code owner, and the buffer is per-thread, so overlapping sets across workers are realistic — especially since a rolled-back transaction leaves ids in the buffer for an unrelated later request to flush (see the buffer comment).

Cheap fix — make the lock order deterministic:

```python
PythonCode.objects.select_for_update()
    .filter(id__in=ids)
    .order_by("id")
    .values_list("id", flat=True)

VitaliiPovoroznykHYS (changes requested):

VitaliiPovoroznykHYS (approved):

ArturDrohomeretskyiHYS (approved):

Issue #45, opened 2025-09-05
Fix hardcoded similarity_threshold and search_limit in human input


OleksandrKutsevHYS:
Fix hardcoded similarity_threshold and search_limit in human input

PR #765, opened 2026-08-20
chore(EST-3868-BE): switch yaml.FullLoader to yaml.safe_load

OleksandrKutsevHYS:
switch yaml.FullLoader to yaml.safe_load

FullLoader can resolve YAML tags into arbitrary already-imported Python objects (e.g. python/name:os.system), which is unnecessary risk for files that only ever carry plain scalar env values. safe_load rejects those tags outright while parsing identical legitimate configs unchanged.

Security review finding #45 (MEDIUM).


VitaliiPovoroznykHYS (approved):

OleksandrKutsevHYS:

PR #769, opened 2026-08-20
fix/EST-3869 no security headers and public swagger

OlgaMageriaHYS:
### 1. Security headers and TLS at nginx

The edge server block set no security response headers. The SPA is proxied
straight from the frontend container and never passes through Django's
middleware, so it carried no `X-Frame-Options` at all and was framable.

### 2. `ENABLE_API_DOCS` flag

`/api/schema/`, `/swagger/` and `/redoc/` were served to anonymous callers
(drf-spectacular's `SERVE_PERMISSIONS` defaults to `AllowAny`), advertising the
full API surface.
Defaults to `True`, so dev and existing deployments are unchanged. The production
manifest ships it `False`.

## Checklist

### DIFF
--- a/src/django_app/django_app/settings.py
+++ b/src/django_app/django_app/settings.py
@@ -434,13 +434,6 @@ SCHEDULE_CHANNEL = os.environ.get("SCHEDULE_CHANNEL", "schedule_channel")
 WEBHOOK_HOST_NAME = os.getenv("WEBHOOK_HOST_NAME", "localhost")
 WEBHOOK_PORT = int(os.getenv("WEBHOOK_PORT", 8009))
 
-ENABLE_API_DOCS = os.getenv("ENABLE_API_DOCS", "True").lower() in (
-    "true",
-    "1",
-    "yes",
-    "on",
-)
-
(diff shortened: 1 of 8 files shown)


src/django_app/django_app/settings.py
```diff
@@ -434,6 +434,13 @@ def _resolve_log_level() -> str:
 WEBHOOK_HOST_NAME = os.getenv("WEBHOOK_HOST_NAME", "localhost")
 WEBHOOK_PORT = int(os.getenv("WEBHOOK_PORT", 8009))
 
+ENABLE_API_DOCS = os.getenv("ENABLE_API_DOCS", "True").lower() in (
```
VitaliiPovoroznykHYS: I don't think we need separate flag for this. There is already DEBUG flag, if DEBUG=False that means we don't show swagger and other docs endpoints

VitaliiPovoroznykHYS (changes requested):

ArturDrohomeretskyiHYS (approved):

VitaliiPovoroznykHYS (approved):

PR #778, opened 2026-08-21
fix(EST-0000): resolve FlowAssistant API key off the async path

igorpolishchukhys:
LiteLLMClient._build_kwargs resolved the LLMConfig's api_key_secret with a synchronous Secret.objects lookup, and it did so from inside FlowAssistantService.stream_reply — an async generator driven by the SSE view under ASGI (gunicorn + UvicornWorker). Django 5.1 guards the DB cursor with @async_unsafe, so that query raised SynchronousOnlyOperation and killed the stream on the first turn of any conversation whose LLMConfig had a secret attached. It only appeared to work when no secret was set, because SecretResolver.resolve returns early on secret_id=None.

Introduced by c43e66d49, which replaced a plain `cfg.api_key` attribute read with a resolver call — turning a free attribute access into a DB round-trip without moving it off the async path.

The caller now resolves the credential via sync_to_async and passes the plaintext in, so LiteLLMClient no longer touches the ORM at all — correct layering for a transport client, and it keeps the query out of the stream. api_key is keyword-only with no default: None stays a legitimate explicit value for providers needing no credential, but an omission at a call site fails loudly at construction instead of silently sending an unauthenticated request. The plaintext's lifetime is unchanged — it lives only on the per-request client instance, exactly as kwargs["api_key"] already did.

The stream view gains an explicit SecretResolutionError branch so an unreadable credential (rotated SECRET_KEY, missing row) reports itself instead of falling through to the catch-all "Unexpected error."

Adds the regression test that was missing. Every other streaming test in test_flow_assistant.py patches get_llm_client with a MagicMock, so _build_kwargs never ran; the one test that did call it was synchronous, so no event loop existed and the async-unsafe guard never tripped. The new test is @pytest.mark.asyncio, drives the real client with a real Secret row, and stubs only litellm.acompletion. Verified to fail against the pre-fix code.

Note: the two reformatted hunks in service.py are pre-existing ruff-format debt on main, not intentional changes.

### DIFF
--- a/src/django_app/tables/services/flow_assistant/service.py
+++ b/src/django_app/tables/services/flow_assistant/service.py
@@ -22,7 +22,11 @@ from django.utils import timezone
 
 from utils.logger import logger
 
-from tables.exceptions import LLMConfigMissingError, LLMConfigInvalidError, ToolExecutionError
+from tables.exceptions import (
+    LLMConfigMissingError,
+    LLMConfigInvalidError,
+    ToolExecutionError,
+)
 from tables.models.flow_assistant_models import (
     FlowAssistant,
     FlowAssistantConversation,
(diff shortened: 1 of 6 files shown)


igorpolishchukhys:

ArturDrohomeretskyiHYS (approved):

PR #791, opened 2026-08-27
fix(EST-2752-FE): fix stale waypoint causing zigzag/loop connections

ViraKalenykHYS:
## Description

Implementation of task [EST-2752](https://hys-enterprise.atlassian.net/browse/EST-2752).
### Resolves two reported connection-routing problems in the flow editor:
1. Connections randomly rendering as visual loops/zigzags
2. Connections sometimes routing under (behind) other nodes
### Root causes & fixes
- **Stale waypoint reuse**: after a node drag, the reroute logic was reusing outdated/overly complex waypoints instead of recomputing a clean path, even when a simple straight line was already available. Fixed the early-return and fallback logic in `computeSegmentAvoidanceWaypoints` to correctly collapse back to a clean route once a detour is no longer needed.
- **Connections cutting through nodes**: a vertical-stack route (right-port source directly above a left-port target) could draw its first leg straight through the source node's own body when no other node was blocking the direct path. Extended `buildForwardVerticalStackRoute` to also route over the top in this case.
- **Classification-flip race with realtime collaboration**: when a connection's backward/forward classification flipped (e.g. after a drag), a WebSocket echo of the flow state could reset the tracking used to detect that flip before the reroute pass had a chance to clean up the old (semantically different) waypoint, permanently stranding a backward-arc "route height" hint as a literal segment bend point. Replaced the silent tracking resync with a full reroute pass so this can no longer race.
- **Classification decision table node overlap**: added correct collision bounds for `CLASSIFICATION_TABLE` nodes so other nodes/connections no longer overlap them.

## Checklist

### DIFF
--- a/frontend/src/app/visual-programming/core/helpers/helpers.spec.ts
+++ b/frontend/src/app/visual-programming/core/helpers/helpers.spec.ts
@@ -85,4 +85,40 @@ describe('isBackwardConnection', () => {
 
         expect(isBackwardConnection(conn, [source])).toBe(false);
     });
+
+    it('stays true for a horizontally-backward connection regardless of the vertical gap between nodes', () => {
+        const source = node('source', 100, 300, 330, 60, 'source_out-right', 'right');
+        const conn = connection('source', 'source_out-right', 'target', 'target_in-left');
+
+        for (const dy of [5, 25, 40, 60, 120, 200]) {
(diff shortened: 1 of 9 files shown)


ViraKalenykHYS:

BohdanSydirHYS (approved):

PR #811, opened 2026-09-01
feat(EST-3939): Webhook/Telegram/Twilio inbound auth: trigger-level strategy pattern

NikitaZalohaHYS:
## Summary

Replaces the old per-node webhook auth (`WebhookNodeAuth`, DB-configurable HMAC/static-header schemes) with a fixed Strategy pattern attached at the `WebhookTrigger` level:

- **Webhook-trigger strategy** — constant `EPICSTAFF_API_KEY` header, user-settable secret via the existing `Secret` model.
- **Telegram strategy** — Telegram's own `X-Telegram-Bot-Api-Secret-Token` mechanism, also user-settable.
- **Twilio strategy** — existing `X-Twilio-Signature` check, secret always mirrors `TwilioChannel.auth_token_secret`; `kind=twilio` can also be reserved on a bare trigger ahead of a `TwilioChannel` being created.

A `WebhookTrigger` now serves exactly one node type — cross-type attachment (e.g. a `TelegramTriggerNode` on a trigger already claimed by a `WebhookTriggerNode`) is rejected, and a trigger's auth kind is automatically cleared once the last node of that kind is detached, making it reassignable. Telegram/Twilio auth is rejected on localhost-only triggers, since those aren't externally reachable.

Also included:
- The webhook ingress now fails closed (401) when no valid auth config resolves for a path, instead of the previous fail-open default.
- Twilio's signature validation now resolves its callback URL from the registered tunnel domain instead of the caller-controlled `X-Forwarded-Host`/`X-Forwarded-Proto` headers, closing a URL-binding gap; the matching nginx `/voice` location fix.
- `WEBHOOK_PORT` default aligned between `docker-compose.yaml` and the webhook service's own settings.
- A failed Telegram re-registration after a secret change no longer discards the saved secret or raises an uncaught error — it surfaces as a warning on the response instead.

**Breaking changes**: any existing webhook-trigger integration computing an HMAC signature must switch to sending the plain `EPICSTAFF_API_KEY` header. A `WebhookTrigger` can no longer drive both a `WebhookTriggerNode` and a `TelegramTriggerNode` simultaneously (previously supported).

**Scope note**: this PR is backend-only. Frontend changes for this work exist locally but are intentionally not included here.


src/webhook/app/controllers/webhook_routes.py
```diff
@@ -83,138 +41,36 @@ async def handle_webhook(
...
-    elif matched_principal is not None:
-        # Already logged above (unauthenticated-fallback forwarding).
-        pass
-    else:
-        logger.info(
+    authenticated = (
+        bool(token)
+        and bool(auth.secret)
+        and hmac.compare_digest(token, auth.secret)
+    )
```
VitaliiPovoroznykHYS, line 58: `compare_digest` on `str` operands requires both to be ASCII-only

There is potential risk having non-ascii symbols

src/django_app/tables/services/webhook_trigger_service.py
VitaliiPovoroznykHYS, line 147: `_convert_trigger_auth` now calls `secret_resolver.resolve`, which raises `SecretResolutionError` if the database row is gone or the value fails to decrypt. Because this call sits inside a list comprehension building the single `WebhookConfigData` payload (with no `try/except` per config), **one bad trigger in any org** will crash the entire payload generation.
> 
> *Impact:* `_re_register_webhooks` will just log the error and push nothing. As a result, webhook config changes for *all* orgs will silently stop landing. If the webhook service restarts, every path will be unregistered, and the `/register-webhooks/` endpoint will hard 500. (The old `_convert_node_auth` read plaintext columns and could not fail, so this is a newly introduced failure mode).
> 
> *Fix:* Replace the list comprehension with a standard `for` loop and wrap the `secret_resolver.resolve` call in a `try/except` block per config. This ensures a single undecryptable secret only skips that specific trigger instead of taking down the entire system.

src/django_app/tables/signals/webhook_signals.py
```diff
@@ -127,3 +100,78 @@ def webhook_trigger_node_post_delete_handler(sender, instance: WebhookTriggerNod
...
+    trigger = instance.webhook_trigger
+    if trigger is None:
+        return
+
+    existing = getattr(trigger, "auth", None)
+    if existing is not None and existing.kind != WebhookTriggerAuthKind.TWILIO:
+        raise ValidationError(
+            f"Cannot sync Twilio auth onto trigger {trigger.pk}: its auth is "
+            f"already configured for kind='{existing.kind}'."
+        )
```
VitaliiPovoroznykHYS, line 162: **Bug:** Raising `django.core.exceptions.ValidationError` from a `post_save` signal causes two major issues. First, DRF only translates `rest_framework.exceptions.ValidationError`, so this surfaces as an unhandled 500 error, not a 400. Second, it fires *after* the row is already written to the DB. Since `ATOMIC_REQUESTS` isn't enabled and the serializer save isn't wrapped, the `TwilioChannel` row is permanently committed while the client sees a crash.

**Fix:** `TwilioChannelSerializer.validate` (`realtime_serializers.py:244`) already rejects this correctly. Either move this specific logic to `TwilioChannel.clean()`, or change this signal to log-and-skip (leaving the auth row untouched) rather than raising an exception.

VitaliiPovoroznykHYS (changes requested):
**Security Risk (Brute-forceable Secrets):** The new ingress relies on a plain bearer-token equality check, but there is no minimum length or entropy enforced for either kind. 

`kind=webhook` has no validator at all, and `kind=telegram` allows a length of `{1,256}`. Because `/webhooks/{path}` is not rate-limited, a 1-character secret is easily brute-forceable. *Recommendation:* Enforce a minimum length (e.g., ≥32 characters) inside `set_trigger_auth_secret`.

Also, update the docs

VitaliiPovoroznykHYS (approved):

PR #819, opened 2026-09-01
fix(EST-3962): RBAC role-gate bypass fix + removal of unused register-* endpoints

NikitaZalohaHYS:
## Summary

**EST-3962** — `RealtimeAgentChatViewSet` and `ConversationRecordingViewSet` declared `rbac_resource_type = ResourceType.VOICE` but used `IsAuthenticatedOrApiKey`, which does zero role/bitmask checking. Any org member (or a self-issued API key), regardless of seeded role, could update/delete conversation recordings and delete realtime agent chat history within their own org. Org scoping itself was already correct (`OrgScopedChildViewSetMixin`) -- this closes the missing role-axis gate by wiring in `HasOrgPermission`, matching the pattern used by ~60 other RBAC-declaring viewsets in the same file. Verified the SYSTEM API-key path still works (`SystemServicePrincipal.is_superadmin = True` short-circuits `HasOrgPermission` before org-context resolution).

**EST-3964** (subtask of EST-3962) -- found via a follow-up sweep of the whole realtime/voip/webhook surface for the same declared-but-unwired pattern (none remained). Instead, found two worse, cross-tenant issues with zero RBAC/org-scoping at all:
- `RegisterTelegramTriggerApiView` -- unfiltered global pk lookup let any user force `setWebhook` against another org's Telegram bot.
- `RegisterWebhooksApiView` -- no auth beyond `IsAuthenticated`; republished every org's webhook configs (including auth secrets) to Redis on any authenticated call.

Neither endpoint has any real caller in this codebase (frontend, other services, or management commands), so both were removed entirely rather than gated:
- `register-webhooks/` is superseded by the existing `manage.py register_webhooks` command (shell/deploy access only).
- `register-telegram-trigger/`'s normal resync already happens automatically via `telegram_signals.py` on node changes.

The underlying service methods (`WebhookTriggerService.register_webhooks`, `TelegramTriggerService.register_telegram_trigger`) are untouched -- still called internally by signal handlers and the management command.

## Breaking changes
- `POST /api/register-telegram-trigger/` and `POST /api/register-webhooks/` no longer exist. No known caller in this codebase.
- Viewer/Member-role users can no longer update/delete `ConversationRecording` or delete `RealtimeAgentChat` in their own org (this was the vulnerability; only Org Admin retains write/delete, per the seeded VOICE bitmask).

## Test plan
- [x] New tests: 12 tests covering Viewer/Member denial + Org Admin regression on both viewsets (`test_rbac_voice_chat_recording_permissions.py`)
- [x] 3 pre-existing org-scoping tests adjusted to use an org-admin fixture (Member no longer has CREATE on VOICE); added explicit role-denial test
- [x] `manage.py check` clean after endpoint removal; direct regression suite (`telegram_trigger_node_api_test.py`) passes in full
- [x] `backend-code-reviewer` pass on both EST-3962 and EST-3964 diffs: no Critical/Concern findings
- [x] `architect` sweep confirmed zero remaining declared-but-unwired RBAC gaps across the full realtime/voip/webhook surface
- [x] Manual live verification against a running environment with Viewer/Member accounts (not yet performed)

Note: one pre-existing, unrelated test failure was found and left untouched per project convention (a `token_budget` kwarg drift in `SessionManagerService`'s test stubs, unrelated to this diff) -- flagged separately for triage.


docs/telegram_trigger_node/Telegram_Trigger_Node_Endpoints.md
```diff
@@ -75,15 +75,29 @@ the Webhook Developer Guide's "Webhook Inbound Authentication
 2. Register Webhooks (Global Sync)
 ----------------------------------
 
-- **Endpoint**: `/api/register-webhooks/`  
-- **Method**: `POST`
+There is no HTTP endpoint for this anymore -- `/api/register-webhooks/` and
+`/api/register-telegram-trigger/` were removed (no real callers existed
+beyond the tests written for a since-abandoned auth-gating fix). Webhook/
+tunnel registration still happens automatically:
```
ArturDrohomeretskyiHYS: we do not need to support backwards compatibility. Docs must represent current state of the code base

src/django_app/tables/views/model_view_sets.py
ArturDrohomeretskyiHYS: Restrict Update and Patch actions, instead of provide full CRUD via viewset

ArturDrohomeretskyiHYS (changes requested):

ArturDrohomeretskyiHYS (approved):

PR #822, opened 2026-09-02
fix/EST-3941 org scope LLM/Embedder models

ViraKalenykHYS:
## Description

Organizations could affect each other through the four provider model tables
(`LLMModel`, `EmbeddingModel`, `RealtimeModel`, `RealtimeTranscriptionModel`).
Rows are either shared built-ins owned by no organization, or an org's own
custom rows — and four things broke that boundary.

### Fixes

**Built-ins were writable by anyone.** The hybrid queryset includes built-ins so
they stay *readable*, and the same queryset backs `get_object`. The realtime
registries had no guard at all; for LLM/embedding only `name` and `predefined`
were protected, so a member could repoint a built-in's `base_url` and redirect
every tenant's inference traffic. Now 403 `built_in_model_immutable` on any
write to an `org IS NULL` row — including for superadmin, since built-ins are
seeded by `upload_models`, never the API.

**Custom rows could escape into the shared bucket.** `is_custom` was writable,
so `PATCH {"is_custom": false}` kept the row's `org` but made it visible to
every tenant. Both ownership flags are now server-controlled.

**Name uniqueness was global.** One org's private model name blocked every other
org — a cross-tenant DoS and an existence leak. Now per-org, plus a partial
constraint keeping built-ins unique among themselves.

**Quickstart, seeding and import all wrote unstamped rows.** Each now stamps the
resolved org. Seeding additionally could crash for the whole deployment
(`MultipleObjectsReturned`) from one org admin's POST, and an empty catalog JSON
would have deleted all 537 shared rows.

### Migrations

`0235` adopts quickstart-orphaned rows into the org that references them (rows
claimed by 0 or 2+ orgs stay shared; nothing is deleted) and collapses realtime
duplicates. `0236` swaps the constraints. Data and DDL are split deliberately —
Postgres won't `ALTER TABLE` after DML on it in the same transaction.

### Testing

56 new tests. Verified end-to-end on a throwaway DB against every row shape both
migrations can hit (18/18). No new failures against the pre-existing baseline.

## Checklist


src/django_app/tables/serializers/model_serializers/embedding_serializers.py
```diff
@@ -25,8 +26,28 @@ class EmbeddingModelSerializer(TagHandlingMixin, serializers.ModelSerializer):
 
     class Meta:
         model = EmbeddingModel
-        fields = "__all__"
-        read_only_fields = ["org", "created_by"]
+        # Explicit rather than "__all__" — see LLMModelSerializer.
```
ArturDrohomeretskyiHYS: remove comment

src/django_app/tests/migration_tests/test_adopt_orphan_provider_models.py
ArturDrohomeretskyiHYS, line 1: remove this test

ArturDrohomeretskyiHYS (changes requested):

ArturDrohomeretskyiHYS (approved):

PR #825, opened 2026-09-03
Fix/est 3937 django settings

BogdanTimchenkoHYS:
### What
- Add a shared, typed environment loader (Env) plus human-readable duration/size parsers.
- Make every backend service read its configuration from the environment only — in-code defaults removed, and variable names and casts unified across services.
- Rework docker-compose.yaml so each service is explicitly handed every variable it consumes.

### Why
Configuration was split between django-environ, ad-hoc os.getenv, and per-service default values that had silently drifted from what Docker actually injected; making the environment the single source of truth surfaces a missing or mismatched variable at container start instead of deep in runtime.

### How
- `Env` reads `os.environ` (and a `.env` only when not in Docker) through typed getters: `int/float/time/byte_size` raise on a malformed value, `str/list` never raise, and `bool` maps anything outside `{1,y,yes,true,on}` to `False` — a wrong numeric/duration cast fails loudly, but a `bool` over a non-true/false value is silently `False` (worth attention in review).
- With defaults gone, each read is required at its call site, so the compose blocks now enumerate every consumed variable as `${VAR:?}`; a service not handed a variable it reads fails at import.
- Durations/sizes are written human-readably (`5s,` `50mb`) and parsed by the new helpers, so several settings moved from `int` to `time`/`byte_size`.
- Redis clients passed `user=` to redis-py (which wants `username=`); only the redis-py call changed — the wrapper keeps its user constructor param.
- crewdb's healthcheck now authenticates as the configured superuser/database instead of a hardcoded `postgres`, and the Redis container provisions an ACL user matching `REDIS_USER`.

### Side Effects
- Blast radius is every service: a variable a service reads that the environment doesn't provide now takes it down at startup, where a default previously masked it — one renamed/missing variable = one dead service.
- Local / non-Docker runs must now supply a complete environment; anything relying on old fallbacks breaks.
- Cast changes alter failure modes: a malformed value that used to pass as a string can now raise (`int`/`float`/`time`); a `bool` set to anything other than true/false silently becomes False.
- crewdb uses the superuser the data volume was first initialized with; an existing volume created under a different `DB_USER` mismatches the new healthcheck/ACL assumptions.
- All Redis-using services depend on that ACL user existing with the matching password; otherwise they fail to connect.
- nginx resolves upstreams at startup and hard-fails if any of frontend/realtime/webhook/django is down — restarting a single upstream can leave nginx flapping until it's back.

### Relates
[EST-3937](https://hys-enterprise.atlassian.net/browse/EST-3937)

## Checklist


ArturDrohomeretskyiHYS:
merge main and solve conflicts again

src/crewdb/custom-docker-entrypoint.sh
```diff
@@ -274,6 +304,30 @@ EOF
...
+        RAISE NOTICE '[crew] CRUD granted on tables_memorydatabase';
+    ELSE
+        RAISE NOTICE '[crew] tables_memorydatabase not found, skipping';
+    END IF;
+END
+\$\$;
+EOF
+    echo "[crew] Permissions granted"
+}
+
```
ArturDrohomeretskyiHYS: deprecate, as we dont need access to DB for crew container

src/crewdb/custom-docker-entrypoint.sh
```diff
@@ -113,16 +113,46 @@ EOF
...
+\$\$;
+
+GRANT USAGE ON SCHEMA public TO "${crew_user}";
+
+ALTER DEFAULT PRIVILEGES IN SCHEMA public
+    REVOKE ALL ON TABLES FROM "${crew_user}";
+EOF
+    echo "[crew] Role ready: ${crew_user}"
+}
+
```
ArturDrohomeretskyiHYS: deprecate, as we dont need access to DB for crew container

ArturDrohomeretskyiHYS (changes requested):
merge main again

ArturDrohomeretskyiHYS (approved):

PR #834, opened 2026-09-07
fix(ETS-4003): fix resize block

oleksandrponomarenko-hash:
## Checklist


frontend/src/app/shared/directives/resizable-section.directive.ts
```diff
@@ -0,0 +1,149 @@
...
+            this.renderer.listen(this.el.nativeElement, 'pointerdown', (event: PointerEvent) =>
+                this.onResizeStart(event)
+            );
+        });
+    }
+
+    private onResizeStart(event: PointerEvent): void {
+        this.isResizing = true;
+        this.startY = event.clientY;
+        this.startHeight = this.sectionTarget().getBoundingClientRect().height;
```
ViraKalenykHYS, line 67: A plain click (no movement) on the handle reads startHeight from a collapsed target as 0 and immediately commits that as the new saved height on pointerup. Could we skip the commit when the pointer never actually moved?

frontend/src/app/features/agent-definitions/pages/agent-definitions-page/components/explorer/explorer.component.html
```diff
@@ -79,11 +83,22 @@
                             (attachSurface)="onAttachSharedSurface($event)"
                         />
                     </div>
+                    @if (!shouldFillBody('agents')) {
+                        <div
+                            class="explorer__section-resize-handle"
+                            [appResizableSection]="sectionHeightKey('agents')"
```
ViraKalenykHYS, line 89: sectionMinHeight here defaults to 80px, which isn't a multiple of the 28px tree-row height — this reintroduces the row-cut-off issue this PR fixes for the fill body (100→168). Mind bumping it to 168 (or a multiple of 28) here too?

frontend/src/app/features/agent-definitions/pages/agent-definitions-page/components/explorer/explorer.component.scss
```diff
@@ -148,10 +149,39 @@
...
+    min-height: 168px;
+    overflow-y: auto;
+}
+
+.explorer__section-body--resized {
+    flex-shrink: 0;
     overflow-y: auto;
 }
 
+.explorer__section-resize-handle {
```
ViraKalenykHYS, line 164: The handle's only visual feedback is :hover/:active, which drops out as soon as the cursor moves off this 6px strip during an actual drag — even though the resize keeps working underneath. Could we add a real isDragging state (like ColumnResizeDividerComponent already has) bound to a persistent class, and maybe widen the hit area a bit?

frontend/src/app/shared/directives/resizable-section.directive.ts
```diff
@@ -0,0 +1,149 @@
...
+                    this.storageKey(),
+                    this.pendingHeight,
+                    this.sectionMinHeight(),
+                    this.effectiveMaxHeight
+                );
+            });
+        });
+    }
+
+    private onPointerEnd(commit: boolean): void {
```
ViraKalenykHYS, line 167: onPointerEnd commits the height without cancelling/flushing a still-pending rAF from the last pointermove — a fast release could commit a stale position. ColumnResizeDividerComponent.stopDragging() already guards against this — could we reuse that approach?

frontend/src/app/shared/services/section-height.service.ts
```diff
@@ -0,0 +1,57 @@
...
+    private getOrCreateSignal(key: string): WritableSignal<number | null> {
+        let existing = this.heightSignals.get(key);
+        if (!existing) {
+            existing = signal(this.readStoredHeight(key));
+            this.heightSignals.set(key, existing);
+        }
+        return existing;
+    }
+
+    private readStoredHeight(key: string): number | null {
```
ViraKalenykHYS: Unlike SidebarWidthService.readStoredWidth, the stored height isn't clamped against min/max on load. Could we clamp here too, in case the container shrinks between sessions?

frontend/src/app/features/agent-definitions/pages/agent-definitions-page/components/explorer/explorer.component.ts
```diff
@@ -162,6 +172,71 @@ export class ExplorerComponent {
...
+                ) as HTMLElement | null;
+                if (!sectionEl) continue;
+                const sectionHeight = sectionEl.getBoundingClientRect().height;
+
+                if (this.shouldFillBody(id)) {
+                    // Currently absorbing leftover space — only its guaranteed floor must survive.
+                    const headerHeight =
+                        (sectionEl.firstElementChild as HTMLElement | null)?.getBoundingClientRect().height ?? 0;
+                    reserved += headerHeight + FILL_BODY_MIN_HEIGHT;
+                } else if (id === 'storage') {
```
ViraKalenykHYS: The id === 'storage' check is a bit fragile since ExplorerSectionId already has an unused knowledge variant. Could this be derived from position in sectionOrder instead, so it still works if the order ever changes?

frontend/src/app/shared/directives/resizable-section.directive.ts
```diff
@@ -0,0 +1,149 @@
+import { Directive, ElementRef, inject, input, NgZone, OnDestroy, OnInit, output, Renderer2 } from '@angular/core';
```
ViraKalenykHYS, line 1: This is now a third hand-rolled drag-resize implementation in shared/ (after ResizableSidebarDirective and ColumnResizeDividerComponent), without the Escape-cancel/keyboard-resize the column divider already supports. Any chance of reusing that one instead?

frontend/src/app/shared/services/section-height.service.ts
```diff
@@ -0,0 +1,57 @@
...
+        }
+        if (raw == null) {
+            return null;
+        }
+        const parsed = Number(raw);
+        return Number.isFinite(parsed) ? parsed : null;
+    }
+
+    private clamp(px: number, min: number, max: number): number {
+        return Math.min(max, Math.max(min, Math.round(px)));
```
ViraKalenykHYS: clamp() returns below the floor when max < min. Minor edge case, but worth a Math.max(min, ...) guard?

frontend/src/app/features/agent-definitions/pages/agent-definitions-page/components/explorer/explorer.component.ts
```diff
@@ -27,6 +28,12 @@ import { StorageSectionComponent } from './storage-section/storage-section.compo
 import { SurfacesSectionComponent } from './surfaces-section/surfaces-section.component';
 
 const SIDEBAR_STORAGE_KEY = 'agents';
+/**
+ * Matches .explorer__section-body--fill's CSS min-height — the floor a resize drag must never push
+ * the filling section below. 168 = 6 full tree rows (tree-node.component.scss: .row min-height 28px),
+ * so the filling section never bottoms out mid-row (100px only fit ~3.5 rows, cutting the 4th item off).
+ */
+const FILL_BODY_MIN_HEIGHT = 168;
```
ViraKalenykHYS: FILL_BODY_MIN_HEIGHT is duplicated by hand as a raw 168px in the .scss, kept in sync only by a comment. Could this be a shared constant/CSS var instead?

frontend/src/app/features/agent-definitions/pages/agent-definitions-page/components/explorer/explorer.component.html
```diff
@@ -79,11 +83,22 @@
                             (attachSurface)="onAttachSharedSurface($event)"
                         />
                     </div>
+                    @if (!shouldFillBody('agents')) {
```
ViraKalenykHYS, line 86: This resize-handle block is pasted twice, identical apart from the section id/label, right next to the declarative EXPLORER_SECTIONS array. Worth a @for instead?

frontend/src/app/features/agent-definitions/pages/agent-definitions-page/components/explorer/explorer.component.html
```diff
@@ -79,11 +83,22 @@
...
+                        <div
+                            class="explorer__section-resize-handle"
+                            [appResizableSection]="sectionHeightKey('agents')"
+                            [sectionTarget]="agentsBody"
+                            [sectionMaxHeightFn]="sectionMaxHeightFn('agents', agentsBody)"
```
ViraKalenykHYS, line 92: sectionMaxHeightFn(...) allocates a new closure on every change-detection run even though it's only read once at drag start. Could this be memoized per section id?

ViraKalenykHYS (changes requested):

frontend/src/app/features/agent-definitions/pages/agent-definitions-page/components/explorer/explorer.component.html
oleksandrponomarenko-hash: But I add 84px because it`s good for UI

frontend/src/app/features/agent-definitions/pages/agent-definitions-page/components/explorer/explorer.component.ts
oleksandrponomarenko-hash: I didn't add this because it breaks the app.

frontend/src/app/shared/directives/resizable-section.directive.ts
oleksandrponomarenko-hash: Full reuse of ColumnResizeDividerComponent is not suitable here.

frontend/src/app/features/agent-definitions/pages/agent-definitions-page/components/explorer/explorer.component.ts
oleksandrponomarenko-hash: But I add 84px because it`s good for UI

frontend/src/app/features/agent-definitions/pages/agent-definitions-page/components/explorer/explorer.component.html
oleksandrponomarenko-hash: This is not necessary for this.

frontend/src/app/features/agent-definitions/pages/agent-definitions-page/components/explorer/explorer.component.ts
ViraKalenykHYS: Thanks for clarifying about the 84px — that's fine. The question was actually a bit different: not about the number itself, but that it's hand-duplicated in two places (FILL_BODY_MIN_HEIGHT in .ts and min-height: 84px in .scss), kept in sync only by a comment. Could the .scss read this from a CSS var instead, so they can't drift apart in the future?

ViraKalenykHYS (approved):

PR #837, opened 2026-09-07
fix(EST-4012): enforce is_active on realtime channel lookup, invalidate realtime cache

NikitaZalohaHYS:
## Summary
- `RealtimeChannel.is_active` was never enforced on the inbound voice-channel lookup path — an operator setting a channel inactive as a containment action had no effect. Fixed at both call sites (`RealtimeChannelViewSet.lookup_by_token`, `TwilioService.configure_webhook`); an inactive channel now 404s identically to an unknown token (no existence leak).
- Even with that fixed, the realtime service cached the lookup response for up to 60s with no invalidation. Added a new `realtime_channels:invalidate` Redis channel: Django publishes on `RealtimeChannel` `post_save`/`post_delete`, and the realtime service's existing Redis listener now also subscribes to it and pops just the affected channel's cache entry immediately. The 60s TTL stays as a defense-in-depth fallback for missed pub/sub messages.
- Also fixes a pre-existing `docker-compose.yaml` bug (unrelated to the above, found while verifying this fix live): the `crewdb` service passed `MANAGER_DB_USER`/`KNOWLEDGE_DB_USER`/`REALTIME_DB_USER` to the container, but its init script expects `DB_MANAGER_USER`/`DB_KNOWLEDGE_USER`/`DB_REALTIME_USER` — the name mismatch silently skipped those three role creations, which broke the `auto_grant_table_permissions()` trigger (and therefore every migration) on any fresh bring-up.


src/django_app/tables/signals/webhook_signals.py
VitaliiPovoroznykHYS, line 1: Can we please avoid using signals. They are unreliable, create unnecessary coupling and create a lot of confusion when debugging.

Move this logic to overridden `save()` and `delete()` method on `RealtimeChannel` model

src/django_app/tables/services/twilio_service.py
```diff
@@ -159,10 +159,10 @@ def configure_webhook(self, phone_sid: str, channel_token: str, org_id: int) ->
         try:
             channel = RealtimeChannel.objects.select_related(
                 "twilio__webhook_trigger__ngrok", "twilio__webhook_trigger__localhost"
-            ).get(token=token)
+            ).get(token=token, is_active=True)
```
VitaliiPovoroznykHYS: This should not be filtered in this way. We need to use custom manager, so we don't forget to use is_active=True all the time. 

I recommend to inherit RealtimeChannel from SoftDeleteFields and then we can remove filter from here

src/django_app/tables/views/model_view_sets.py
```diff
@@ -1671,7 +1671,7 @@ def lookup_by_token(self, request):
                 "twilio__webhook_trigger__ngrok",
                 "twilio__webhook_trigger__localhost",
             )
-            .filter(token=token)
+            .filter(token=token, is_active=True)
```
VitaliiPovoroznykHYS: Same here as [this comment](https://github.com/EpicStaff/EpicStaff/pull/837/changes#r3956576597)

VitaliiPovoroznykHYS (changes requested):

src/django_app/tables/signals/webhook_signals.py
NikitaZalohaHYS: Removed signals and moved their logic to overrided `save()` and `delete()`

src/django_app/tables/services/twilio_service.py
NikitaZalohaHYS: I didn't use SoftDelete. I added new manager for Active fields. SoftDelete not exactly what there is needed. So now we have two managers: one for SoftDelete instances that we can hide, if they was soft deleted; and one for Active entities that could be turned off but not deleted

src/django_app/tables/views/model_view_sets.py
NikitaZalohaHYS: Same here as [this comment](https://github.com/EpicStaff/EpicStaff/pull/837#discussion_r3957419831)

VitaliiPovoroznykHYS (approved):

PR #841, opened 2026-09-08
Feature/est 3999 admin panel api keys

ArturDrohomeretskyiHYS:
## Checklist


ElenaPrygunHYS (approved):

src/django_app/tables/services/rbac/permission_catalog.py
```diff
@@ -20,6 +24,8 @@
     # {"code": "list", "label": "List", "bit": int(Permission.LIST)},
 ]
 
+GRANTABLE_ACTION_BITS: int = reduce(or_, (entry["bit"] for entry in ACTION_METADATA), 0)
```
VitaliiPovoroznykHYS: **Org Admins cannot assign the Viewer role**
- **The Bug:** The `GRANTABLE_ACTION_BITS` mask calculates to 95 instead of 31 because it fails to exclude the `USE` bit.
- **The Impact:** When an Org Admin tries to assign the Viewer role, the permission comparison fails (`66 & ~31 = 64`). The assignment is blocked (403 error), and Viewer disappears from the assignable-roles list for non-superadmins.
- **Fix:** Split the constant. Keep `ACTION_METADATA` as-is for the frontend, but define the ceiling mask explicitly as `CREATE|READ|UPDATE|DELETE|EXPORT`.

src/django_app/tables/services/rbac/effective_permissions.py
```diff
@@ -35,6 +38,48 @@ def can(self, resource_type: str, action: Permission) -> bool:
...
+        (the bits the assigned role grants) so the two cannot drift. A
+        resource absent from this principal's map counts as zero, so requesting
+        nothing on a resource is never an escalation. Superadmin covers
+        everything.
+
+        Only `GRANTABLE_ACTION_BITS` are compared. USE and LIST are not in the
+        catalog, are rejected by role validation and are read by no code, but
+        they do occur in the built-in seeds -- comparing them would let dead
+        data refuse a legitimate grant (Org Admin, `flows: 31`, could not
+        assign Viewer, `flows: 66`).
```
VitaliiPovoroznykHYS: Related to the `GRANTABLE_ACTION_BITS`: this docstring says "USE and
LIST are not in the catalog," but `USE` **is** in `ACTION_METADATA` and so is
in `GRANTABLE_ACTION_BITS`. Only `LIST` is actually excluded.

src/django_app/tables/services/rbac/role_management_service.py
```diff
@@ -3,109 +3,489 @@
...
-            return role.org_id
-        if role.is_built_in and role.name == BuiltInRole.SUPERADMIN:
-            return None
-        return context_org_id
+    def _assert_name_available(org_id, name, exclude_role_id) -> None:
+        clash = Role.objects.filter(org_id=org_id, name__iexact=name)
+        if exclude_role_id is not None:
+            clash = clash.exclude(pk=exclude_role_id)
+        if clash.exists():
+            raise RoleNameConflictError()
```
VitaliiPovoroznykHYS, line 478: **Race condition allows duplicate role names in the same org**
- **The Bug:** Role creation checks `.exists()` and `.create()` separately without a database-level uniqueness constraint. 
- **The Impact:** Concurrent requests or double-clicks will create duplicate roles, breaking the per-org name uniqueness that `RoleNameConflictError` expects.
- **Location:** `role_management_service.py:473-478`, `role.py:20-21`
- **Fix:** Add a `UniqueConstraint` on `(org, lowercased_name)` to the database and catch the `IntegrityError`.

src/django_app/tables/services/rbac/membership_management_service.py
```diff
@@ -0,0 +1,273 @@
...
+            actor, membership.org_id, action=Permission.DELETE
+        )  # no-leak 404
+        self._assert_not_self(actor, membership)
+        self.assert_can(effective, Permission.DELETE)
+        membership.delete()
+        logger.info(
+            "MembershipManagementService.remove_member actor={a} membership={m}",
+            a=getattr(actor, "email", "system"),
+            m=membership_id,
+        )
```
VitaliiPovoroznykHYS, line 215: **Orgs can be left with zero admins (and it's untested)**
- **The Bug:** The "last-org-admin" protection was removed. 
- **The Impact:** An Org Admin can remove the *last other* admin in their org, leaving the organization unmanageable without superadmin intervention. Furthermore, deleting a custom role bulk-reassigns users to Viewer without checking if they were the only admins.
- **Fix:** Either restore the `last_org_admin` guard, or explicitly document this "superadmin rescue" behavior and write a test for the non-superadmin removal path.

VitaliiPovoroznykHYS (approved):

src/django_app/tables/services/rbac/membership_management_service.py
ArturDrohomeretskyiHYS: it is false assumption:

An Org Admin can remove the last other admin in their org - every org admin (or person who can manage memberships and roles) can remove or decrease permission of other user in Organization. BUT he never can change own role or permission or remove himself from org. it is restricted. 
So it means Org Admin (or person who can manage memberships and roles) always will stay in org, except the Superadmin will remove him or downgrade his permissions - and it is accepted behavior for us


**Need to fix:**

Furthermore, deleting a custom role bulk-reassigns users to Viewer without checking if they were the only admins. -
need to restrict deleting caller's own role

PR #886, opened 2026-09-15
EST-3798: Stamp and export the acting user on session/interaction records

NikitaZalohaHYS:


src/django_app/tables/migrations/0245_backfill_session_principal.py
VitaliiPovoroznykHYS, line 1: We don't really need to backfill existing records, there is already default and no reason to fill them in current state

VitaliiPovoroznykHYS (changes requested):

src/django_app/tables/migrations/0245_backfill_session_principal.py
NikitaZalohaHYS: Backfill migration file is deleted now

VitaliiPovoroznykHYS (approved):

PR #888, opened 2026-09-16
Fix/est 3840 web pages and documents can hijack your agents

PavloNesterukHYS:
## Description

1. Tool result envelope (src/agent/) — all tool output sent to the LLM is now wrapped in a JSON envelope at a single point in the agent loop (agent_loop.py:378), after the emitter call so the UI still receives raw content. Built with json.dumps, so injected text stays in an escaped string slot and cannot close its field or forge sibling keys like a fake "note": "this content is trusted". Previously only knowledge search was protected; MCP, python-code and all 75 catalog tools arrived as bare strings. One chokepoint covers every executor, every failure path, and any future or user-copied tool.

2. SSRF guard in scraper tools (src/shared/tools/) — added to the 4 scraping tools (scheme allow-list, DNS resolution, rejection of private/loopback/link-local/reserved/multicast addresses). Also closed a redirect bypass: they followed redirects by default, so a guarded public URL could 302 → http://127.0.0.1/ and be fetched unguarded. Redirects are now walked manually, re-guarded on every hop, capped at 3, with cross-host hops refused. Stops an injection-steered agent reaching Redis, Postgres, MinIO or 169.254.169.254. Known limit: DNS-rebinding TOCTOU remains.

3. Description sanitizer (src/django_app/) — user-authored tool descriptions are sent to the model in the tool schema every turn, so they were a writable prompt slot needing no external content. Now stripped of C0 control characters and capped at 1024 on create, update, and copy (the copy service bypasses the serializer).

Also: deleted two tests broken since they were committed — one imported a module that never existed, the other targeted a factory deleted

## Checklist


VitaliiPovoroznykHYS (changes requested):
SSRF function DRY

VitaliiPovoroznykHYS (approved):

PR #909, opened 2026-09-17
Fix/est 4145 rbac race

BogdanTimchenkoHYS:


src/django_app/tables/services/rbac/user_management_service.py
```diff
@@ -170,52 +171,71 @@ def revoke_superadmin(self, actor, target_user_id):
...
             "UserManagementService.revoke_superadmin actor={a} target={t}",
             a=getattr(actor, "email", "system"),
             t=target.email,
         )
+
         return target
 
     @transaction.atomic
-    def set_user_active(self, actor, target_user_id, is_active):
+    def set_user_active(self, actor, target_user_id, value):
```
ArturDrohomeretskyiHYS, line 205: why renamed is_active -> value?

src/django_app/tables/services/rbac/user_management_service.py
```diff
@@ -170,52 +171,71 @@ def revoke_superadmin(self, actor, target_user_id):
...
+        superadmins_map = {sa.pk: sa for sa in superadmins}
+
+        if target_user_id in superadmins_map:
+            if len(superadmins_map) <= 1:
+                raise LastSuperadminError()
+            target = superadmins_map[target_user_id]
+        else:
+            target = UserModel.objects.select_for_update().filter(pk=target_user_id).first()
+            if target is None:
+                raise UserNotFoundError()
```
ArturDrohomeretskyiHYS, line 189: maybe better to keep this functionality in UserManagementGuards?

src/django_app/tables/services/rbac/user_management_service.py
BogdanTimchenkoHYS: it is more readable.

src/django_app/tables/services/rbac/user_management_service.py
BogdanTimchenkoHYS: Yes, but in current implementation breaks `UserManagementGuards` consistency.
UserManagementGuards is checker of user invariant, by contract other methods it returns nothing or boolean and raises exceptions if error occurs. I didn't thought up it.

ArturDrohomeretskyiHYS (approved):

PR #916, opened 2026-09-18
fix(EST-4073-BE): make webhook trigger path globally unique

NikitaZalohaHYS:
## Summary
- Closes a HIGH-severity cross-org availability bug: `WebhookTrigger.path` was only unique per-org, so any invited Member of one org could import a graph declaring a webhook path already used by another org — both orgs then hit `AmbiguousWebhookPathError` (409) on inbound requests to that path, since the downstream `webhook` ingress service resolves by bare path with no org discriminant.
- `path` is now globally unique on its own (DB-level constraint), enforced consistently via a shared `validate_path_uniqueness()` helper at all three creation paths: the direct `/api/webhook-triggers/` serializer, the graph-import serializer (the actually-exploited path via `POST /api/graphs/import/`), and the live nested serializer.
- Migration `0246` includes a safe pre-migration step that resolves any pre-existing cross-org (or now cross-provider-type) path collisions by renaming the newer row (`<path>-org<id>`, with collision/length-overflow guards), logging each rename.
- Corrected a false docstring claim in `tunnel_registry.py` that the path was "cryptographically unguessable" — it's a plain user-chosen string, now protected by the DB constraint instead.


src/django_app/tables/serializers/base_serializers.py
```diff
@@ -263,25 +263,23 @@ def get_current_secret_reference(self, source):
         return existing.secret if existing is not None else None
 
     def validate(self, data):
+        from tables.services.webhook_trigger_service import validate_path_uniqueness
```
VitaliiPovoroznykHYS: move import to the top or is there any reason to import here?

VitaliiPovoroznykHYS (changes requested):

src/django_app/tables/serializers/base_serializers.py
NikitaZalohaHYS: moved

VitaliiPovoroznykHYS (approved):

